PR #25, opened 2017-04-10
added new params:

randaz81:
* fullscalePWM: 800 for BLL, 1333 for mc4, 32000 for ems, 3360 for mc4plus
* ampsToSensor: 1000 ofr all boards

to be propagated to all robots!


randaz81:
FYI: @valegagge @davidetome @ale-git

davidetome:
in which files we have to add these params?

PR #28, opened 2017-04-20
Update right_arm_calib to avoid pulley rewinding

S-Dafarra:
This avoid the rewind of the right index distal pulley during calibration. This issue was due to a wrong reading of motor encoders (see https://github.com/robotology/icub-support/issues/418), thus this is a workaround to avoid https://github.com/robotology/icub-support/issues/447 .


S-Dafarra:
About the base branch I agree, it should be ``devel``. Simply I was not informed about the purpose of the ``devel`` branch.
Regarding the first issue you raised, I would say that the answer is in the middle. Indeed the original value seemed to be out of the reachable ranges of the finger (thus is a yes), BUT, if the motor encoders were working fine, this issue would have never occurred (thus it may be a no too). The problem is that if the [issue](https://github.com/robotology/icub-support/issues/418) on right index persists, the finger would be compromised since the startup, while now it loses a fraction of degree when completely extended. Given that in the recent future the ``.local`` folder may be easily deleted, I would have preferred to have this value somewhere :grin:

PR #29, opened 2017-04-20
Remove repetition of TORQUE_CONTROL parameters

pi-q:
TORQUE_CONTROL parameters were set twice in iCubGenova02/hardware/motorControl/right_upper_arm-ems3-mc.xml. Would it be better if we keep only one set of parameters?


traversaro:
Thanks @pi-q ! 

@davidetome It seems a problem that affects other robots as well, see for example: https://github.com/robotology/robots-configuration/blob/593d1d170896f14baea38a1795c7ff90af0bad0b/iCubNottingham01/hardware/motorControl/right_upper_arm-ems3-mc.xml#L63 .

PR #31, opened 2017-05-29
Modifications of iCubGenova04 configurations files.

S-Dafarra:
Can you please check whether everything is coherent with the new naming system (@valegagge) and whether I forgot something (@nunoguedelha @gabrielenava @DanielePucci @pi-q @traversaro @francesco-romano).

### DIFF
--- a/iCubGenova04/calibrators/head-calib.xml
+++ b/iCubGenova04/calibrators/head-calib.xml
@@ -21,14 +21,14 @@
         <group name="CALIBRATION">
         <param name="calibrationType">      3           3           3           3           5           5           </param>
         <param name="calibration1">   	    0	        0           0	        0	    3000        3000	    </param>
-        <param name="calibration2">         0           0           0	        0           0   	0	    </param>
+        <param name="calibration2">         0           0           0	        0           8192   	8192        </param>
         <param name="calibration3">     14362.08    19967.08    63342.92        58462.92    0    	0           </param>
 
         <param name="calibration4">         0           0           0           0            0          0           </param>
         <param name="calibration5">         0           0           0           0            0          0           </param>        
         
-        <param name="calibrationZero">      180.0       -180.0      180.0       180.0       45.0        45.0        </param>
-        <param name="calibrationDelta">     0.9         0.1         4.6         -9.0        5.0        10.0         </param>
+        <param name="calibrationZero">      180.0       -180.0      180.0       180.0       0           0           </param>
+        <param name="calibrationDelta">     0.9         0.1         4.6         -9.0        -5.0        -10.0       </param>
        
         <param name="startupPosition">      0.0         0.0         0.0         0.0         0.0         0.0         </param>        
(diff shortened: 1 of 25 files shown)


S-Dafarra:
I suggest to add `?w=1` in the URL to avoid seeing the whitespace differences :smiley:

valegagge:
Hi @nunoguedelha ,

we are working to have a unique default configuration for all robots, but for now it is not possible because each robot is different from another.

I think you have not clear the idea behind PID_DEFAULT and PID_CUSTOM: in PID_DEFAULT there are pid values for right behaviour of the robot. In case two robots are equals, than their PID_DEFAULT are equal. So, the values of pid in tag PID_DEFAULT are tunned for that type of mechanics/eletronic.

In case the user wants to perform an experiment that need particular values of pid, the user should define PID_CUSTOM tag, and use it by setting PID_CUSTOM in the tag CONTROL to perform the experiment.
When the experiment finished, the user restores in CONTROL tag the PID_DEFAULT. 
When he/she wants to perform again the experment, he/she needs to set PID_CUSTOM in tag CONTROL only.

In other words, PID_CUSTOM doesn't contain value spefic for green robot, but values specific for your experiment.

In this way you can commit your custom pid, without demage default configuration.

nunoguedelha:
Hi @valegagge , I understand better now the intention of this process, thanks.

So what would happen if we have multiple different PID configurations tuned for different experiments/demos? Probably we would start cumulating PID config blocks, increasing maintenance effort, considering, among other troubles, that visualising the differences between configurations would be inefficient. This would obviously aggravated by the tuning of other additional parameters.

I had a former experience in managing configuration/calibration files across product families and versions (actually config/calibration of RFICs across phone models), using this same type of configuration handling, and after a while it didn't end well. The complexity and maintenance effort went on growing, and we ended up trying new solutions.

I suggest we use the **GIT branching and labelling** (git tags) to handle different versions of PID configurations or any other type of config parameters (there will be others). This will allow easy comparisons between configs, easy merge from default config updates to derived specific configs. And meanwhile, an "install" of the checkedout version would copy the selected files to the `.local` folder, from where the user could try any personal unofficial change.
I would say it's worth a try to analyse that solution..

valegagge:
Hi @nunoguedelha ,

a user is free to use this feature or to continue to work as usual.

I doesn't think we'll  comulate a lot PID_CUSTOM block.  Each research group how many experiments does? Do all experiments need a different pid group?

More over, in the future will be possible to choose different pid algorithm type, so in some case it is necessary different pid values.

At the end,I think it's better if we would have a chat about this.

cheers 
Valentina

valegagge:
hi @S-Dafarra ,

please rebase your branch on last commit of devel and than you can merge. 
Please use option "no fast forward" in merge command in order to create a merge commit.

thanks!

S-Dafarra:
I have just ``rebased`` over ``devel``, but I cannot merge the pull request.

PR #32, opened 2017-06-06
Update calibration config files

barbalberto:
Update wrapper by adding 'Calibrator' tag to allow remote calibration from motorgui
Merged together calib file for arm and hand


pattacini:
Too old PR, closing.
Anyway we have a backup.

PR #48, opened 2017-10-27
Updated iCubGenova04 arms, legs, torso and neck joints calibration offsets

nunoguedelha:
Calibrated all the offsets manually, using the level. Also fixed left knee position PID integration param and left and right hip roll joint limits.

### DIFF
--- a/iCubGenova04/calibrators/head-calib.xml
+++ b/iCubGenova04/calibrators/head-calib.xml
@@ -16,8 +16,8 @@
             <param name="velocityHome">     10          10          10          10          10          10         </param>
         </group>
 
-        <!-- joint logical number               0           1           2           3           4           5           -->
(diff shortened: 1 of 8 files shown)


nunoguedelha:
New update of the joint offsets.

nunoguedelha:
@valegagge @traversaro could one of you approve this pull request please? Thanks

traversaro:
Thanks!

valegagge:
Hi,
I have an option and an information to give you
My option is about the big values of calibrationDekta,it should value few degree. So there is somethig wrong.

The information I would give is is about ne calibration type for joint with absolute encoder.
I implemented a new version of calibration 3, called calibration 12, that semplifies founding out of calibration parameters. Calibration 12 let the user to take raw encoder at zero position and just to write it in parameter1;the firmware will calculate all other parameters. Obviusly the calibration delta remains with the same meaning, while calibration zero should be 0. I think we should change calibration type 3 in 12 asap,but i know that  you have a damo on monday so if you prefer we can make thi change next week.

PR #59, opened 2017-12-18
Update torso calibration

nunoguedelha:
Update the joint encoder offsets on the torso (manual calibration). Also checked the neck joints calibration which didn't require any update.


nunoguedelha:
Sure, I keep that in mind. Thanks!

PR #62, opened 2018-02-14
Update config files for iCubGenova04

S-Dafarra:
Some modifications to the configuration files, in particular of the additional imu configuration which was creating problems with the IMU Bosh.

The calibration of the shoulder have been changed after a slide of the bearings (which happened because of a fall).


iCubGenova04/estimators/wholebodydynamics.xml
```diff
@@ -38,30 +38,32 @@
...
+                   ,3.127489e-03,-1.180096e-02,1.690016e-04,8.666081e-01,2.382600e-02,-6.542111e-02 
+                   ,-1.245428e-19,1.921199e-18,-2.013402e-18,-1.928427e-16,1,1.655828e-18 
+                   ,0.0,0.0,0.0,0.0,0.0,1.0)</param> 
+         </group>
 
 
         <group name="WBD_OUTPUT_EXTERNAL_WRENCH_PORTS">
 	    <param name="/wholeBodyDynamics/left_arm/cartesianEndEffectorWrench:o">(l_hand,l_hand_dh_frame,root_link)</param>
             <param name="/wholeBodyDynamics/right_arm/cartesianEndEffectorWrench:o">(r_hand,r_hand_dh_frame,root_link)</param>
-            <param name="/wholeBodyDynamics/left_leg/cartesianEndEffectorWrench:o">(l_foot,l_sole,root_link)</param>
```
traversaro: This is a change w.r.t. to all the other robots, and the semantics of the same port in the legacy `wholeBodyDynamics` module. Can we revert it?

S-Dafarra: What would you like to revert? The body of which we compute the external force or its full definition? I still need the measurements provided by the *_foot port, but on the corresponding *_sole frame. If we need to keep both we may think about a better name. 
Just as a side comment I don't think that a modification wrt other robots is a big issue considering that we are in a robot specific folder.

traversaro: Sorry for the late reply. Let's merge it as it is the current de-facto situation on the robot, and I will open another issue for this.

traversaro:
If it is not used anymore on the robot, can you delete the branch @S-Dafarra ? Thanks!

S-Dafarra:
> Once this PR gets merged, should I keep pushing on this branch or directly on ``devel``?

If the answer is "push directly on ``devel``" then I will delete it. In my opinion it may be better to keep this branch active and regularly merge it on ``devel``, but I am open to suggestions.

traversaro:
Ah sorry, I did not get this. 
I am agnostic to this, if you like doing that do it.

S-Dafarra:
I just wanted to avoid polluting the ``devel`` branch while keeping a full history of commits. For the time being I would keep this branch then. If we realize the name is ugly or not understandable we can plan to change strategy.

PR #68, opened 2018-05-17
Align configuration files to YARP devel

S-Dafarra:
Modifies the iCubGenova04 configuration files to be used with the latest version of YARP and iCub.


S-Dafarra:
@valegagge @randaz81 Do you think we can merge this?

valegagge:
I think it is ok!
I have one only suggestion: in order to avoid to increase number of files, I suggest to delete old file abouts strain and rename file for strain2 in strain.

S-Dafarra:
> I have one only suggestion: in order to avoid to increase number of files, I suggest to delete old file abouts strain and rename file for strain2 in strain.

We are still testing those new features. For the time being I think it is better to keep both files and remove them in a next PR.

PR #78, opened 2018-08-28
Fix Gyroscopes full-scale configuration to get the expected default FS of 250 deg/s

nunoguedelha:
Fix Gyroscopes full-scale configuration through the parameter ETH:X in order to get the expected default FS = 250 deg/s.

| ETH | CTRL_REG4 FS1-FS0 | Full-scale |
| --- | --- | ---------- |
|  0  | 01  | 500 deg/s |
|  1  | 00  | 250 deg/s |
|  2  | 01  | 500 deg/s |
|  3  | 10  | 2000 deg/s |

Refer to [l3gd20 datasheet](https://www.st.com/resource/en/datasheet/l3gd20.pdf) and https://github.com/robotology/icub-main/pull/543#issuecomment-416299969.

### DIFF
--- a/iCubGenova02/hardware/inertials/left_arm-eb2-j4_15-inertials.xml
+++ b/iCubGenova02/hardware/inertials/left_arm-eb2-j4_15-inertials.xml
@@ -40,7 +40,7 @@
                                             eoas_accel_mtb_int      eoas_accel_mtb_int      eoas_accel_mtb_int      eoas_gyros_mtb_ext        
                     </param>
 
-                    <param name="location"> ETH:0                   ETH:0                     
+                    <param name="location"> ETH:0                   ETH:1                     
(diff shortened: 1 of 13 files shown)


traversaro:
@randaz81 Just to understand, how is this interacting with the new  automatically generated files for robots-configuration ?

valegagge:
I see now, that we merged in master. This is an error!!

Since I 'm working on the final solution and it wil be available in few days, I don't revert this merge.

I'll revert this changes after I'll merged the final solution.

nunoguedelha:
@valegagge we were targetting master since this was a direct dependency of https://github.com/robotology/icub-main/pull/543. We were initially considering both PRs as a small fix, not a feature, for having homogeneous units across all inertial sensors (strain2 IMUs, MTB accelerometers, etc).

But since it's temporary, you're right, it makes sense to merge to devel only.

PR #87, opened 2018-12-18
recalibrated robot after re-install

arrenglover:
iCubGenova02 has been re-calibrated after the re-installation of the head.


pattacini:
Hi @arrenglover

This is definitely a stale PR that I'm aiming to close.
Given all the major overhaul that's happening, I don't believe it'd be worth being considered.

pattacini:
In any case, this branch can be recovered from the back up available at https://github.com/robotology-legacy/robots-configuration/tree/iCubGenova02_head_recalibration.

PR #89, opened 2019-02-06
Update left_arm-eb24-j4_7-skinSpec.xml

maggia80:
Added a special config for the MTB board of the skin of the hand. The configuration should be skin-type 1 for the hand.


maggia80:
yes, you are right! thanks

PR #91, opened 2019-03-12
Genova02eyelids

marcoaccame:


davidetome:
Hi Marco,

 referrimng to easyrdmine task# 6018, this early morning Emiliano checked the 0B7_M1 motor wiring (eyelids) on iCubGenova02, and noticied that the wiring is wrong.

So @julijenv please do the wiring as described in the above mentioned task and then @marcoaccame have to reverse again the PID sign of the motor.

thanks you all,

Davide

@maggia80 @Fabrizio69

julijenv:
Hi @marcoaccame ,
I double checked, the PID can be changed as previously agreed.
You can proceed to commit this change.

marcoaccame:
ok, i fixed w/ https://github.com/robotology/robots-configuration/pull/91/commits/6e62e1bc674c5124298beb20c52b7ebf543f23dc

PR #92, opened 2019-03-13
Remove leftover install of non-existant camera directory in vizzy files

traversaro:
Fix leftover of https://github.com/robotology/robots-configuration/pull/86 .


traversaro:
cc @plinioMoreno

plinioMoreno:
Vizzy's new cameras are using ROS interface, so we don't need that folder anymore

traversaro:
Great, thanks!

PR #112, opened 2019-08-09
iCubGenova04: arms, legs and torso joint offsets recalibration

nunoguedelha:
These changes update the joint offsets calibration:
- left and right arms
- Recalibrated left and right legs after r_hip_yaw encoder replacement
- Recalibrated the torso after ironbot tests

These offsets have been used for at least a couple of months on the Green iCub(Genova04), and run on all the demos.


nunoguedelha:
Ciao @S-Dafarra , this is just the last branch that was being used for some time, not yet integrated in `devel`. After this PR is merged, `devel` and `devel_iCubGenova04` won't differ, except for minor changes in the Current Control config not impacting the behavior of the controllers.
Then, from the resulting `devel` commit, I can create a merge back to `devel_iCubGenova04`, such that this `devel` commit will be the new "base" for future merges from or to `devel_iCubGenova04`.

CC @traversaro

S-Dafarra:
Just a side comment. Why some ``calibrationZero`` changed?

julijenv:
the robot torso has been removed and when it has been remounted it has been mounted 90 degree away from before unmounting.

nunoguedelha:
If I recall well it was more or less the same case for the hip yaw.

S-Dafarra:
Got it! Thanks!

PR #115, opened 2019-08-21
Add the configuration files in order to use the robot without the foreArms.

FabioBergonti:
Add the configuration files in order to use the robot without the forearms.
Basically, I've copied the configuration files of iCubGenova04, then I removed/commented the lines where the forearms joints were called.

All the files have been called with the suffix iRonCub.

The files that I've created are:
- `iCubGenova04/calibrators/left_arm-calib_iRonCub.xml`
- `iCubGenova04/calibrators/right_arm-calib_iRonCub.xml`
- `iCubGenova04/estimators/wholebodydynamics_iRonCub.xml`
- `iCubGenova04/iRonCub-noLegs_wbd.xml`
- `iCubGenova04/iRonCub_wbd.xml`
- `iCubGenova04/wrappers/motorControl/left_arm-mc_wrapper_iRonCub.xml`
- `iCubGenova04/wrappers/motorControl/right_arm-mc_wrapper_iRonCub.xml`

I've modified the `calibrators` because it was necessary to remove the columns of the pronosupinations and the wrists.

I've modified the `wrappers/motorControl` because it was necessary to change the number of joints  and to comment some lines of unused joints

I've created two startup files:
- `iCubGenova04/iRonCub_wbd.xml` (for iCub without the forearms)
- `iCubGenova04/iRonCub-noLegs_wbd.xml` (for iCub without the torso, without the forearms and without the legs)

Note that `iCubGenova04/iRonCub-noLegs_wbd.xml` is not compatible with `iCubGenova04/estimators/wholebodydynamics_iRonCub.xml` because the latter recalls leg/torso joints.

cc @nunoguedelha @gabrielenava

### DIFF
--- a/iCubGenova04/calibrators/left_arm-calib_iRonCub.xml
+++ b/iCubGenova04/calibrators/left_arm-calib_iRonCub.xml
@@ -6,31 +6,31 @@
                 <xi:include href="../general.xml" />
 
 		<group name="GENERAL">
-		    <param name="joints">16</param> <!-- the number of joints of the robot part -->
+		    <param name="joints">4</param> <!-- the number of joints of the robot part -->
 		    <param name="deviceName"> Left_Arm_Calibrator </param>
 		</group>
 
-<!-- joint logical number              0         1           2         3           4      5      6     7     8     9    10    11    12     13     14    15 -->
- 
+<!-- joint logical number              0         1           2         3 	-->
+
 <group name="HOME">
-<param name="positionHome">           -30        30         0          45           0     0      0    35    65     0     0     0     0      0      0     0 </param>
-<param name="velocityHome">           10         10         10         10           30    30     30   60    30    30    30    30    30     30     30    30  </param>
-</group>                                                                        
-                                                                                
-<group name="CALIBRATION">                                                      
-<param name="calibrationType">        3          3          3          3            5     12     12       7     7     6     6     6     6      6      6     6 </param>
-<param name="calibration1">           0   	 0          0          0        -1500    38300  25230     0     0     0     0     0     0      0      0     0 </param>
-<param name="calibration2">	      0          0          0          0	16384     0      0        0     0  9102  9102  9102  9102   9102   9102  3640 </param>
-<param name="calibration3">         30431.08  3135.14    6479.08     64526.92       0     0      0        0     0    -1     1    -1     1     -1     1     -1 </param>
-<param name="calibration4">           0          0          0          0            0     0      0     2130   200   241   499   254   465    226    485  694  </param>
-<param name="calibration5">           0          0          0          0            0     0      0     2530  2372   46    22     9   10     33      5   135  </param>
-<param name="calibrationZero">        -180.00   -315.00    180.00     -180.00       0     0      0        0     0     0     0     0     0      0      0     0 </param>
-<param name="calibrationDelta">       106.4      -9.6      -17.0       -0.8         0     0      0        0     0     0     0     0     0      0      0     0 </param>
-                                                                                
-<param name="startupPosition">        -35.97     29.97      0.06       50.00        0     0      0    25    65     0     0     0     0      0      0     0       </param>
-<param name="startupVelocity">        10.0       10.0       10.0       10.0        30    30     30    60    100   100   100   100    100   100    100  100       </param>
-<param name="startupMaxPwm">          5000       5000       5000       5000      2000     0     0      0     0    0       0     0     0      0      0     0      </param>
-<param name="startupPosThreshold">    2          2          2          2           90   90     90      90   90    90     90    90    90     90     90    90      </param>
+<param name="positionHome">           -30        30         0          45	</param>
+<param name="velocityHome">           10         10         10         10	</param>
+</group>
+
+<group name="CALIBRATION">
+<param name="calibrationType">        3          3          3          3	</param>
+<param name="calibration1">           0   	 0          0          0	</param>
+<param name="calibration2">	      0          0          0          0	</param>
+<param name="calibration3">         30431.08  3135.14    6479.08     64526.92	</param>
+<param name="calibration4">           0          0          0          0	</param>
+<param name="calibration5">           0          0          0          0	</param>
+<param name="calibrationZero">        -180.00   -315.00    180.00     -180.00	</param>
+<param name="calibrationDelta">       106.4      -9.6      -17.0       -0.8	</param>
+
+<param name="startupPosition">        -35.97     29.97      0.06       50.00	</param>
+<param name="startupVelocity">        10.0       10.0       10.0       10.0	</param>
+<param name="startupMaxPwm">          5000       5000       5000       5000	</param>
+<param name="startupPosThreshold">    2          2          2          2	</param>
 </group>
 
 <param name="CALIB_ORDER">(0 1 2 3) </param>
@@ -46,6 +46,3 @@
 		<action phase="interrupt3" level="1" type="abort" />
 
 	</device>
-
-
- 
--- a/iCubGenova04/calibrators/left_arm-calib_iRonCub.xml
+++ b/iCubGenova04/calibrators/left_arm-calib_iRonCub.xml
@@ -1,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!DOCTYPE devices PUBLIC "-//YARP//DTD yarprobotinterface 3.0//EN" "http://www.yarp.it/DTD/yarprobotinterfaceV3.0.dtd">
+
+
+	<device xmlns:xi="http://www.w3.org/2001/XInclude" name="left_arm-calibrator" type="parametricCalibratorEth">
+                <xi:include href="../general.xml" />
+
+		<group name="GENERAL">
+		    <param name="joints">4</param> <!-- the number of joints of the robot part -->
+		    <param name="deviceName"> Left_Arm_Calibrator </param>
+		</group>
+
+<!-- joint logical number              0         1           2         3 	-->
+ 
+<group name="HOME">
+<param name="positionHome">           -30        30         0          45	</param>
+<param name="velocityHome">           10         10         10         10	</param>
+</group>                                                                        
+                                                                                
+<group name="CALIBRATION">                                                      
+<param name="calibrationType">        3          3          3          3	</param>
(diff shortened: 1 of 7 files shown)


nunoguedelha:
Ciao @gabrielenava , just for your information, this pull request is just for additional configuration files that the IronCub guys need systematically when running their tests with the jets mounted.

We are pushing this to the integration branch `devel_iCubGenova04` for avoiding endless rebases and branch skeletons in the repo on the `icub-head` machine.

@FabioBergonti is adding some more information to the PR description.

nunoguedelha:
### Another suggestion regarding the calibration files, for easier maintenance and changes traking
Here are two hypothesis I'm not sure are true...
1. The number of joints `<param name="joints">4</param>` and the joint list `<param name="CALIB_ORDER">(0 1 2 3) </param>` could be enough for selecting a subset of joints, without having to remove the extra columns from the table `<group name="HOME">` and `<group name="CALIBRATION">`. If it's not the case, it should be supported, which would simplify the maintenance of the table, as we could always have the full table intact (and easier to update in a full copy paste from the original file), and just change the selection of joints subset depending on the removed parts.
2. when two config files define redundant parameters, the last file to be loaded overrides the first one. Considering that hypothesis (1) is true, we could then handle the removal of the forearm as follows:
#### calibrators/left_arm-calib.xml
=> unchanged

#### calibrators/left_arm-calib_iRonCub.xml
```
<?xml version="1.0" encoding="UTF-8" ?>
<!DOCTYPE devices PUBLIC "-//YARP//DTD yarprobotinterface 3.0//EN" "http://www.yarp.it/DTD/yarprobotinterfaceV3.0.dtd">

        <device xmlns:xi="http://www.w3.org/2001/XInclude" name="left_arm-calibrator" type="parametricCalibratorEth">
                <xi:include href="../general.xml" />

                <group name="GENERAL">
                    <param name="joints">4</param> <!-- the number of joints of the robot part -->
                    <param name="deviceName"> Left_Arm_Calibrator </param>
                </group>

                <param name="CALIB_ORDER">(0 1 2 3) </param>

                <action phase="startup" level="10" type="calibrate">
                    <param name="target">left_arm-mc_wrapper</param>
                </action>

                <action phase="interrupt1" level="1" type="park">
                    <param name="target">left_arm-mc_wrapper</param>
                </action>

                <action phase="interrupt3" level="1" type="abort" />

        </device>
```

#### in iRonCub_wbd.xml
```
...
<!--  CALIBRATORS -->
<xi:include href="calibrators/left_arm-calib.xml" />
<xi:include href="calibrators/right_arm-calib.xml" />
<xi:include href="calibrators/left_arm-calib_iRonCub.xml" />
<xi:include href="calibrators/right_arm-calib_iRonCub.xml" />
...
```
where `calibrators/left_arm-calib_iRonCub.xml` would override just a few "default" parameters set in `calibrators/left_arm-calib.xml`.

If any joint offsets calibration is done on the iCub in its "iRon" or "normal" configuration, only the `calibrators/left_arm-calib.xml` and `calibrators/left_arm-calib.xml` files would be updated.

@traversaro @valegagge 

CC @randaz81

FabioBergonti:
> @FabioBergonti I suggest you remove the INERTIALS information on the arms (devices, remappers, wrappers) if they are not working since anyway the boards are removed when the forearms are removed.

Thank you @nunoguedelha for the advice.
I've modified the files according to [your suggests](https://github.com/robotology/robots-configuration/pull/115#pullrequestreview-278929316). See [the commit](https://github.com/robotology/robots-configuration/pull/115/commits/c227f2ef685149ce3a912f994823d35336c872d2) with the relative modifications.

I'll wait feedback about [how to handle the calibration files of the arm](https://github.com/robotology/robots-configuration/pull/115#issuecomment-524295062) before proceeding with the changes

traversaro:
For the calibration files I would wait for @valegagge , I have no insight on that without starting inspecting the related source code.

valegagge:
Hi guys,
I suggest you to use the default calibration file of arm and modify only `CALIB_ORDER` parameter removing forearm's joints. It is not necessary to change the number of joints (`joints` parameter) and to remove the columns regarding calibration parameters about them. The parser of calibration file checks if the number of columns is equal to the value of `joints`.

Currently I don't know if

> two config files define redundant parameters, the last file to be loaded overrides the first one. 

Maybe @randaz81 can answer to this question. I had never face up with this problem, but I can check if @randaz81 is not available

FabioBergonti:
Hi guys,
With [this commit](https://github.com/robotology/robots-configuration/pull/115/commits/7c19c6020a1fd32fcd7f5e01aa51364392942766), I've updated the calibration files according to what @valegagge has suggested.
Tell me if other modifications are necessary

nunoguedelha:
LGTM. @FabioBergonti I think it's worth trying what I proposed in the point 2 of https://github.com/robotology/robots-configuration/pull/115#issuecomment-524295062. It would avoid to maintain a branch or duplicated files. It should be quick to test if it works.

FabioBergonti:
Hi guys,
Yesterday I did some tests with the robot.

I tried to launch the interface with the old calibration files (see commit https://github.com/robotology/robots-configuration/pull/115/commits/71888ff2608d712ade3952f8da68edc0ed39429c) and it worked.

Then I tried to do the same, but this time using the calibration files with the matrix with 17 joints (see commit https://github.com/robotology/robots-configuration/pull/115/commits/7c19c6020a1fd32fcd7f5e01aa51364392942766). The yarp logger gave me these errors:
```
ParametricCalibratorEth: Left_Arm_Calibrator : number of joints of device to calibrate ( 4 ) does not match the number of joints in calibrator config file ( 16 )
ParametricCalibratorEth: Right_Arm_Calibrator : number of joints of device to calibrate ( 4 ) does not match the number of joints in calibrator config file ( 16 )
```

I've tried to change the param "joints" (the number of joints of the robot part) from 16 to 4 and the  yarp logger gave me these errors:
```
Left_Arm_Calibrator : invalid number of Calibration1 params 17 4
Right_Arm_Calibrator : invalid number of Calibration1 params 17 4
```

At the end I've tried [@nunoguedelha suggestion](https://github.com/robotology/robots-configuration/pull/115#issuecomment-524295062). Here the [commit](https://github.com/FabioBergonti/robots-configuration/commit/190e68256453db684d8537e90e5f22670fb15f8c) with the modifications. Also in this case yarp logger game some errors:
```
Left_Arm_Calibrator : invalid number of Calibration1 params 0 16
Right_Arm_Calibrator : invalid number of Calibration1 params 0 16
```

nunoguedelha:
Ciao @valegagge , are the failures described by @FabioBergonti expected (first two failures). If we summarize:
1. `joints` = 4, table size=4, `CALIB_ORDER`=4 ---> PASS
2. `joints` = 16, table size=16, `CALIB_ORDER`=4 ---> FAIL
3. `joints` = 4, table size=16, `CALIB_ORDER`=4 ---> FAIL

So apparently the three  parameters have to be aligned. Do you confirm this?

S-Dafarra:
Hi @valegagge! Could you please check the problem here? I would like to merge this ASAP.

pattacini:
Because of #119, I would warmly suggest that you change the base branch of this PR to `cleanup/2019.12.feat-01` instead.

I'm gonna try to reach out to @valegagge, although I'm pretty sure that she won't be responsive 'till next week as she's heavily working for our PI01 event.

In the worst case, as `cleanup/2019.12.feat-01` will be merged tomorrow, most likely, we could keep this PR open and then compare against the next `devel` in the upcoming days.

@S-Dafarra @FabioBergonti @nunoguedelha what do you think?

S-Dafarra:
Thanks @pattacini. I have changed the base branch. 

> I'm gonna try to reach out to @valegagge, although I'm pretty sure that she won't be responsive 'till next week as she's heavily working for our PI01 event.

I understand, I think that 
> In the worst case, as cleanup/2019.12.feat-01 will be merged tomorrow, most likely, we could keep this PR open and then compare against the next devel in the upcoming days.

would be a viable option.

pattacini:
Base branch changed to `devel`.

pattacini:
Let's try to revive this @valegagge

valegagge:
Hi guys!
I'm sorry I answer so late!

I checked the code of `parametricCalibratorEth` to understand in details how it works. So I can recap the following:

1. the solution I suggested works only if the device wrapper has all joints. For example if I don't want to calibrate one or more joints of same body part, I simply exclude them from list in parameter `CALIB_ORDER`: in this way `yarprobotinterface` check if all boards and hardware is ok also the boards and hardware related to joint I want to exclude from calibration. So this solution can be suitable when the robot mounts all hardware and boards, but I don't want to calibrate some joints because they are broken or because I don't want the applications use them.

2. In this PR I noticed that the device wrapper configuration have been modified in order to use only the desired joints. In this case the calibration file need to have the same number of joints and the same number of columns of specified by the wrapper.

Therefore, if you need to exclude some joints, because they are not mounted on robot, you need to change also the calibration files. 
I understand this is not an optimal solution, it should be better don't have multiple copies of the calibration files, but currently is seems not possible. 

I hope I clarify a little this matter.

FabioBergonti:
Hi @valegagge 
Thank you for your clarification.

> Therefore, if you need to exclude some joints, because they are not mounted on robot, you need to change also the calibration files.

With the commit [73cd72f](https://github.com/FabioBergonti/robots-configuration/commit/73cd72f8bd59722bace10295ac8fdbf2af13e0f7) I've restored the [initial configuration](https://github.com/FabioBergonti/robots-configuration/commit/71888ff2608d712ade3952f8da68edc0ed39429c) of the calibration files for the arms.
Now their structure is the following:
- `joints param` = 4
- `table size` = 4
- `CALIB_ORDER` = 4

In September when I tested it worked 

cc @gabrielenava

pattacini:
Thanks @valegagge 👍 
Awaiting @gabrielenava 's final comment on this and then merging.

FabioBergonti:
Since the iRonCub files were created several months ago starting from iCubGenova04, I've compared them with the actual version.
[Here the commit](https://github.com/FabioBergonti/robots-configuration/commit/a1072b68e78242b7dd811157e6578dfb50cabfaf) with the updates

cc @gabrielenava

gabrielenava:
>Awaiting @gabrielenava 's final comment on this and then merging.

I had a look at the files too. They look good and up-to-date with latest modifications. To me we can merge.

PR #116, opened 2019-08-22
Xml files changes in order to adopt MASclient and server also for the imu in the head

Nicogene:
This changes have a circular dependency with https://github.com/robotology/icub-main/pull/592.

How do we handle it? 

TODO:

- [x] Add xml files for the new `imuFilter`
- [x] Test


pattacini:
@Nicogene we'll test both the branches at the same time and then as soon as we're happy we'll merge the PR's, all at once.

traversaro:
Note that this PR removes the `/icub/inertial` port, that is unfortunately quite used even by our own software. For sure we need to communicate this change with our users, but we would also need to update the following icub-main modules:  
* wholeBodyDynamics
* gravityCompensation
* iCubGui

Furthermore, both the iCub_SIM simulator and the Gazebo-based one still use the `/icub/inertial` port. 
Probably a f2f meeting would be helpful to decide how to proceed.

pattacini:
I would very much go for the solution that ensures back-compatibility, as the one @traversaro suggested, at least for a given time period.

Nicogene:
The last two commits(60c96c6 and 25946c4) should guarantee the backward compatibility.
In this way both the `ServerInertial` and the `MultipleAnalogSensorsServer` are attaching the imu, and the port `/icub/inertial` is exposed. The include of the `imuFilter` can be commented out for now from the xml files.

Nicogene:
With a6da043 we should be full backward compatible

Nicogene:
This PR has been successfully tested on `iCubGenova02` with and without `imuFilter`.

Ready to be merged.

pattacini:
Let's uncomment the two lines we discussed today.

Nicogene:
I removed the commit from the PR, the imuFilter is run through `yarprobotinterface`

PR #122, opened 2019-10-22
[WIP] iCubGenova04, iCubGenova02: Implement a static model with Stiction/Stribeck, Coulomb and Viscous components

nunoguedelha:
These changes are part of the implementation of https://github.com/dic-iit/element_torque-control-via-current/issues/81 and a dependency of https://github.com/robotology/icub-firmware/pull/101.

## The main impacts are...
- The implemented friction model includes Stiction, Coulomb and Viscous friction compensation as described in https://github.com/dic-iit/element_torque-control-via-current/issues/81.
- This requires the new input parameters, which already existed in the interface but were either never used in the EMS code, either always set to zero in the config files:
    - `stictionUp`, `stictionDown` ($\text{Nm}$): the static friction respectively for positive and negative joint velocities **[never used in the code]**.
(there are some configs we have to cleanup, that use these parameters though. I will push a new commit with the clwanup after I get the feedback from @marcoaccame and @valegagge and @ale-git on this).
    - `k0` ($\text{Nm}$): Coulomb friction Kc [always set to zero in every configuration].
- I have not changed the units of `Kbemf` (Still in $\text{Nm}/(deg \\, s^{-1})$), but now they represent $K_v$ With respect to the joint velocity instead of the motor velocity, such that the values set in the `yarpmotorgui` interface are larger and easier to tune (order of magnitude 1 instead of 0.0001).
- No change for `Ktau` units.
- !!! ALL THE VALUES ARE NOW POSITIVE !!!

## All positive values...
It is painful and error prone to carry around useless signs that depend on the orientation of the joint and the sign used for the gearbox ratio... even if we can distinguish a pattern in the signs across the `ktau` values for a given part, and we got used to them.
Removing the signs alows to have more portable parameters, considering that they don't change that much between two similar motor, and typically between  left and right legs. This also facilitates the portability of a given joint configuration between the robot and the single-motor testing setup I've been working on lately.

This does not impact the signs used on the gearbox ratios set in the `motorcontrol` configuration files. Actually, consider that the sign previously hard coded on the input `ktau`, is now set from the Gearbox ratio in the EMS controller when the parameters are parsed.

More details o the implementation will be given in https://github.com/robotology/icub-firmware/pull/101.

### DIFF
--- a/iCubGenova02/hardware/motorControl/left_leg-eb6-j0_3-mc.xml
+++ b/iCubGenova02/hardware/motorControl/left_leg-eb6-j0_3-mc.xml
@@ -47,14 +47,14 @@
         <param name="outputType">           pwm                       </param>
         <param name="fbkControlUnits">      metric_units              </param>
         <param name="outputControlUnits">   machine_units             </param>
-        <param name="kp">            -1066.66     2066.66    -711.11    -1066.66    </param>
-        <param name="kd">               0.00        0.00       0.00       0.00      </param>
-        <param name="ki">          -10666.64    14222.18   -7111.09    -1066.64     </param>
-        <param name="maxOutput">        8000        8000       8000       8000      </param>
-        <param name="maxInt">           750        750        750       1000        </param>
-        <param name="stictionUp">          0           0          0          0      </param>
-        <param name="stictionDown">        0           0          0          0      </param>
-        <param name="kff">                 0           0          0          0      </param>
+         <param name="kp">            1066.66     2066.66     711.11     1066.66 </param>
+         <param name="kd">               0.00        0.00       0.00       0.00 </param>
+         <param name="ki">           10666.64    14222.18    7111.09     1066.64 </param>
+         <param name="maxOutput">        8000        8000       8000       8000 </param>
+         <param name="maxInt">           750        750        750       1000 </param>
+         <param name="stictionUp">          0           0          0          0 </param>
+         <param name="stictionDown">        0           0          0          0 </param>
+         <param name="kff">                 0           0          0          0 </param>
     </group>
     
     <!-- default position PID: end -->
@@ -67,19 +67,19 @@
         <param name="outputType">           pwm                 </param>
         <param name="fbkControlUnits">      metric_units        </param>
         <param name="outputControlUnits">   machine_units       </param>
-        <param name="kp">           -200        200          0       -200     </param>
-        <param name="kd">              0          0          0          0     </param>
-        <param name="ki">              0          0          0          0     </param>
-        <param name="maxOutput">    8000       8000       8000       8000     </param>
-        <param name="maxInt">        500        500        500        500     </param>
-        <param name="ko">              0          0          0          0     </param>
-        <param name="stictionUp">      0          0          0          0     </param>
-        <param name="stictionDown">    0          0          0          0     </param>
-        <param name="kff">             1          1          1          1     </param>
-        <param name="kbemf">           0          0          0          0     </param>
-        <param name="filterType">      0          0          0          0     </param>
-        <param name="ktau">            0          0          0          0     </param>
-    </group>
+        <param name="kp">           200        200          0        200     </param>
+        <param name="kd">             0          0          0          0     </param>
+        <param name="ki">             0          0          0          0     </param>
+        <param name="maxOutput">   8000       8000       8000       8000     </param>
+        <param name="maxInt">       500        500        500        500     </param>
+        <param name="ko">             0          0          0          0     </param>
+        <param name="stictionUp">     0          0          0          0     </param>
+        <param name="stictionDown">   0          0          0          0     </param>
+        <param name="kff">            1          1          1          1     </param>
+        <param name="kbemf">          0          0          0          0     </param>
+        <param name="filterType">     0          0          0          0     </param>
+        <param name="ktau">           0          0          0          0      </param>
+     </group>
 
     <group name="2FOC_CUR_CONTROL">
         <param name="controlLaw">           low_lev_current     </param>
--- a/iCubGenova02/hardware/motorControl/right_arm-eb3-j0_3-mc.xml
+++ b/iCubGenova02/hardware/motorControl/right_arm-eb3-j0_3-mc.xml
@@ -47,9 +47,9 @@
         <param name="outputType">           pwm                       </param>
         <param name="fbkControlUnits">      metric_units              </param>
         <param name="outputControlUnits">   machine_units             </param>
-        <param name="kp">                   -711.11   -1066.66    -711.11   -1066.66 </param>
-        <param name="kd">                   0.00       0.00       0.00       0.00 </param>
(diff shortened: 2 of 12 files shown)


nunoguedelha:
I will flash the EMS boards on the legs and torso tomorrow from https://github.com/robotology/icub-firmware-build/tree/feature/impl-static-friction-comp, more precisely https://github.com/robotology/icub-firmware-build/commit/bac7e99ac5ae25024c9e9b8c91a1604019a1f13a, such that the DIC guys can run their tests. Some tuning will be required. I won't be available before Friday afternoon.

julijenv:
Hi @nunoguedelha , I noticed that the changes done were creating the repository iCubShenzhen instead of modifying the existing iCubShenzhen01. is there any motivation to that?

nunoguedelha:
Hi @julijenv , have you checked the diffs with respect to right base branch `devel_iCubGenova04`?
The misunderstanding may come from the fact that I initially created the PR based on `devel` by accident, then immediately changed it to `devel_iCubGenova04`.
Please can you check again what you're seeing on your side and let me know?
Thanks.

nunoguedelha:
> I will flash the EMS boards on the legs and torso tomorrow from https://github.com/robotology/icub-firmware-build/tree/feature/impl-static-friction-comp, more precisely [robotology/icub-firmware-build@bac7e99](https://github.com/robotology/icub-firmware-build/commit/bac7e99ac5ae25024c9e9b8c91a1604019a1f13a), such that the DIC guys can run their tests. Some tuning will be required. I won't be available before Friday afternoon.

@DanielePucci @S-Dafarra , after discussing with @GiulioRomualdi on the walking tests in torque control, I think it would be better to take Friday afternoon to perform the tests on current control and tuning the legs+torso friction parameters. This way I can provide support for the tests.

S-Dafarra:
To be honest, I don't feel very comfortable in changing the default torque PID right now. I would prefer to do extensive tests with the yoga and other torque controllers to make sure we are not introducing a regression. Indeed, the yoga is still a fundamental demo showed to visitors and during expositions, and there will not be people working on it in the immediate future. 
I would rather prefer to keep the old parameters as default for the time being, until we are pretty confident that the demo performances are equal or better. I would love to have a faster way to change the PID to use rather than modifying the configuration files.
My second concern is about the signs. Will the old version be still working?

nunoguedelha:
@S-Dafarra , as we discussed F2F yesterday:
- the group configuration used on the Yoga demo are the `TRQ_PID_TUNE` apriori (to be confirmed).
- In these groups, the viscous friction compensation were all set to zero on the torso, legs and arms.
- This PR does not impact the position control parameters at all, as the sign changes only impact the torque control.
- I have restored all the old PID (Kp, Kd, Ki) parameters in the `TRQ_PID_TUNE` groups.
- Only the `Kbemf` differ. We can restore them to 0 if there are any issues.

## Actions remaining
- Re-check all the robots configurations.
- In YARP parsing, use the absolute value of the PID **Feed-Forward** parameters, for backward compatibitlity, with a Warning if the paraemeter was negative.

nunoguedelha:
### A regression was observed on the torque controller with PWM output while testing the Yoga...
with the commit 4b8017a, and the `icub-firmware` commit https://github.com/robotology/icub-firmware/pull/101/commits/45eb0f8ce25609dcdc73207d47a404caaa188e68:

I had added an extra line in the function updating the `ktau` and the `kbemf`, for setting the `ktau` sign from the gearbox ratio sign. The problem is that the function in question was called before the gearbox ratio parameter was parsed from the configuration files. The default gearbox ratio being = 1, this resulted in a wrong sign for `ktau`.

This issue was not observed if any change of PID configuration was done through `yarpmotorgui` prior running the joints in torque control.

### The issue was fixed...
in https://github.com/robotology/icub-firmware/pull/101/commits/48a699377294171bd27197fe7975a5b2bddfc118 and `robots-configuration`https://github.com/robotology/robots-configuration/pull/122/commits/cae6f4407a4e54f3d564ec6a9725e4540d04683a. The fix consists in mapping the motor control variables and feedback variables through a unique parameter `gearbox2jointOrientConv` in the `Motor` structure holding the whole motor configuration used by the EMS motor controller.

The simpler test revealing the issue is the gravity compensation test https://github.com/robotology/whole-body-controllers/blob/master/controllers/fixed-base-joints-control/jointsControl.mdl, with which we can observe clearly the joint torque inversion while all the joints are oscillating symmetrically w.r.t. the robot's sagittal plane.

### Important impact on the position controllers
After this fix, the position controllers are also impacted by the PR:
**The `ktau`, the friction parameters for the torque controller, and the PID gains for all the controllers are all positive**, independently of the gearbox ratio, i.e. of the orientation of the motor axis versus the orientation of the joint axis.
**This is a functional impact, not a performance impact.** A wrong sign in these parameters will either induce a wrong rotation orientation, or the position controller to diverge (the robot explodes).
So it is easy to detect in case you use the wrong sign!

A Yarp change is ongoing for introducing a safeguard against wrong sign in the input parameters.

S-Dafarra:
@nunoguedelha  needs to check that the required changes are done in YARP/icub-main/icub-firmware to lock the respective SW versions with the flashed firmware.

traversaro:
> @nunoguedelha needs to check that the required changes are done in YARP/icub-main/icub-firmware to lock the respective SW versions with the flashed firmware.
> Also needs to open an issue for the R1 changes (Firmware and robots-configuration).

You know that I am a big fan of GitHub-only development workflow as much as possible, but this modifications are quite extensive and touch a lot of repos and may require a f2f meeting to agree on how to proceed with all the details. I opened a ticket on icub-tech issue tracker to track properly this activity.

pattacini:
Base branch changed to `devel`.

pattacini:
Further, the feature branch has been also rebased over `devel`.

pattacini:
Hi @nunoguedelha 

Many things happened since this PR was opened: we have conflicts to solve.
I believe this PR has something to do with your ongoing work on the FW.
What's your plan?

The very first cleanup would consist of rebasing this PR on top of `devel`.

nunoguedelha:
Hi @pattacini , absolutely, the plan is to do a first rebase the branch ASAP (for avoiding diverging too much), as a first step for the tests and merge as soon as we can go back to IIT.
I'll redo a last rebase right before the tests when the COVID-19 emergency is over.

pattacini:
>I'll redo a last rebase right before the tests when the COVID-19 emergency is over.

Don't expect this will happen anytime soon, unfortunately,
Then, could you rebase the branch before, although you cannot test it of course?

As we said, `robots-configuration` will be receiving new dev and we don't want to let this branch diverge too much.

pattacini:
@nunoguedelha I would close this PR as unmergeable, so fork the branch to keep it safe by your side as I'm going to delete it 😉 

Sorry but I have to take action on this. Then, from your fork, you'll have plenty of time to rebase and start it over.

pattacini:
@nunoguedelha has stored the work in his fork 👍

PR #168, opened 2020-02-13
iTeenGenova01: added graspTheBall application files

ilaria-carlini:
Moved configuration files from docker images to robot configuration because these files are robot dependent.


vtikha:
@ilaria-carlini, great! we will also need to remove ```pf3dTracker.ini``` from the ```iTeenGenova01``` root folder as it is an app related configuration file.

PR #169, opened 2020-02-21
[iCubGenova01] Recalibration of iCub right hand after mechanical repair

xEnVrE:
Recalibration of iCub right hand after mechanical repair


pattacini:
@julijenv I'm gonna merge this if you don't object.

julijenv:
thx @pattacini

PR #170, opened 2020-02-27
update wholebodydynamics configuration files after #116

lrapetti:
The name of the `imu` device changed with #116 and the `wholebodydynamics` configuration files where not updated accordingly.

cc @GiulioRomualdi @julijenv @isorrentino @nunoguedelha


traversaro:
This does not immediately blocks this issue, but it may be worth to notice that this problem  is also related to some other config files: https://github.com/robotology/robots-configuration/search?q=%3Celem+name%3D%22imu%22%3Einertial%3C%2Felem%3E&unscoped_q=%3Celem+name%3D%22imu%22%3Einertial%3C%2Felem%3E .

lrapetti:
@traversaro I should have targeted all of them, do you see any specific file missing?

traversaro:
> 
> 
> @traversaro I should have targeted all of them, do you see any specific file missing?

Ah cool, i just saw the change iCubGenova04, thanks for taking care also for the rest of the robots @lrapetti !

PR #174, opened 2020-03-09
iCubGenova02 update wholebodydynamics with l_sole and r_sole

lrapetti:
Updating configuration file for `wholebodydynamics` estimator adding some useful output ports for whole-body demos.


lrapetti:
This change is not related to an experimental feature as those ports are currently used also for `iCubGenova04`, anyway as you mentioned is not a fix either since the estimator works even without this change (but we are limited in the number of demos we can run).

Having said that, feel free to merge it in the branch you think is more appropriate.

pattacini:
If you confirm that `devel` is not a big deal for you, I'll go with `devel`, otherwise we can stick to `master` not to bother you too much.

lrapetti:
> If you confirm that `devel` is not a big deal for you, I'll go with `devel`, otherwise we can stick to `master` not to bother you too much.

I think `devel` is fine, thanks

PR #176, opened 2020-03-09
Update iCubGenova02 torque control gains

lrapetti:
Trying to use torque control on `iCubGenova02`, we found out that with the current gains configuration the performances were bad, and in particular switching to torque control mode for the torso was cause of unstable behaviour (soon starting oscillating).

Given that the current values look like some "standard" values rather than tuned gains, and all the feedforward parameters were missing, I have performed a tuning of the gains starting from the values of `iCubGenova04` robot. In this way we observed better performances, e.g. in performing gravity compensation. 

I know improvement of torque control, also at the firmware level, are in process. However, for the time being I think we can merge those gains that, even if not yet perfectly tune is reached, allows to use torque control and run some demos on `iCubGenova02`.

CC @DanielePucci @S-Dafarra @nunoguedelha

### DIFF
--- a/iCubGenova02/hardware/motorControl/left_arm-eb1-j0_3-mc.xml
+++ b/iCubGenova02/hardware/motorControl/left_arm-eb1-j0_3-mc.xml
@@ -66,19 +66,19 @@
         <param name="controlLaw">           torque              </param>
         <param name="outputType">           pwm                 </param>
         <param name="fbkControlUnits">      metric_units        </param>
-        <param name="outputControlUnits">   machine_units       </param>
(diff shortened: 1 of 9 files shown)


lrapetti:
@pattacini I am targeting devel in this case, but let me know if you thin `master` is more appropriate in this case.

pattacini:
I'll be happy to merge @lrapetti 
Tell me if you would need to receive at least a positive review first from those you mentioned.

S-Dafarra:
I only took a quick look, but I did not spot anything weird. As far as I am concerned, it looks good to me.

PR #192, opened 2020-08-06
fix/strain2-v208

marcoaccame:
This PR is required to align the xml files to the latest version of the strain2 board which is 2.0.8

The change to 2.0.8 was done only for those xml  files which requested the previous version 2.07, so that the system can be aligned.
Here is the [log change of the new firmware build for strain2](https://github.com/robotology/icub-firmware-build/commit/f11975c9276c855fb5e18ef004c63ae22ecbe5fa)


marcoaccame:
fyi @isorrentino @Gianlucamilani @S-Dafarra

PR #194, opened 2020-08-24
ICubGenova08

Nicogene:
This PR adds `iCubGenova08`


davidetome (approved):

pattacini:
`iCubGenova08` is an upper-body only robot that will be working anchored to a table.
Therefore, we hade to modify startup and parking sequences to avoid any potential collision between the robot's arms and the table.

Eventually, we just reduced the velocity with which the pitch of the torso reaches the calibrated position. By doing that, the arms can calibrate safer and farther from the table.

In the real setup, the distance from the table should be greater than the one we tested.

| Calibration | Parking |
| :---: | :---: |
| output1 | output |

PR #196, opened 2020-09-11
added new frame transform from ground_link and base_link

lnobile:
The new frame transform sets a new frame ground_link that represents the ground height with respect to the base_link frame


pattacini (approved):
Hi @lnobile 
Since this is a new addition I'll merge it in `devel` instead.

PR #208, opened 2020-11-25
Added files for F/T sesnors

davidetome:
this branch has updated xml files used on the robot.
It adds also XML files for F/T sensors in the feet and in the arms 

cc @S-Dafarra @prashanthr05


pattacini (approved):
Thanks a lot @davidetome 🚀

PR #220, opened 2020-12-11
fix fingers

triccyx:
- put the param useMotorSpeedFbk equal to 0
This modification will be executed on files: left_arm-eb26-j12_15-mec.xml, right_arm-eb29-j12_15-mec.xml, left_arm-eb25-j8_11-mec.xml,right_arm-eb28-j8_11-mec.xml.
Only on the folders that: (have pc104.xml file OR have xxx_client.xml)
- update the hands coupling matrices
This modification will be executed on files: right_arm-eb28-j8_11-mec.xml.
Only on the folders that: (have pc104.xml file OR have xxx_client.xml)


pattacini (approved):
Thanks @triccyx 🚀 

These fixes stem from the great job @kouroshD did in https://github.com/robotology/icub-tech-support/issues/1019.

Issue #222, opened 2020-12-14
iCubGenova09 yarprobotinterface throws error "Parameter networks use deprecated syntax"


S-Dafarra:
Every time we run the ``yarprobotinterface`` we get the not fatal error
```
Parameter networks use deprecated syntax
```

It seems it is thrown by https://github.com/robotology/yarp/blob/9ce40a416ce872154c90130a04648eb6923eb536/src/devices/ControlBoardWrapper/ControlBoardWrapper.cpp#L489, but I am not able to spot what causes this issue in the configuration files, nor how to avoid that.

S-Dafarra:
I suppose that lines like 
https://github.com/robotology/robots-configuration/blob/a1605113ce72472c3442a8d72bfbe7d97ec7152b/iCubGenova09/wrappers/motorControl/left_leg-mc_wrapper.xml#L8
should be changed to 
```xml
<elem name="FirstSetOfJoints">  (0  1  0  1) </elem> 
```
but the documentation does not seem to be in this way
https://github.com/robotology/yarp/blob/yarp-3.4/src/devices/ControlBoardWrapper/ControlBoardWrapper.h#L171-L175

Then, I am wondering if this is a problem of ``yarp`` or of the configuration files. This problem was also spotted in https://github.com/robotology/icub-main/issues/661

pattacini:
Thanks @S-Dafarra for reporting

I believe that your analysis is correct 👍🏻 

`(0  1  0  1)` is the new preferred way of passing parameters, whereas we have still `0  1  0  1` available as a fallback (see https://github.com/robotology/yarp/blob/9ce40a416ce872154c90130a04648eb6923eb536/src/devices/ControlBoardWrapper/ControlBoardWrapper.cpp#L490-L493).

As a result, the documentation https://github.com/robotology/yarp/blob/yarp-3.4/src/devices/ControlBoardWrapper/ControlBoardWrapper.h#L171-L175 seems outdated.

S-Dafarra:
Actually, it seems that this error was reintroduced in https://github.com/robotology/yarp/commit/5da1cf3c02d288a24470aad222e2ac127dc370a9 when changing the log mechanism in ``yarp``. Before, that line was commented since ages

S-Dafarra:
Actually, that error was commented out since the first time it was pushed https://github.com/robotology/yarp/commit/d164b21ba83c85cdcd8aafb905d6ec1531439a4f

pattacini:
[...]

PR #227, opened 2020-12-22
fix files

triccyx:
Correct syntax errors. Added parenthesis.

### DIFF
--- a/CER01/wrappers/VFT/left_arm-VFT_wrapper.xml
+++ b/CER01/wrappers/VFT/left_arm-VFT_wrapper.xml
@@ -1,30 +1,23 @@
-<?xml version="1.0" encoding="UTF-8" ?>
+<?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE devices PUBLIC "-//YARP//DTD yarprobotinterface 3.0//EN" "http://www.yarp.it/DTD/yarprobotinterfaceV3.0.dtd">
-
 <devices robot="CER01" build="1">
-	<device xmlns:xi="http://www.w3.org/2001/XInclude" name="left_arm_VFTserver" type="virtualAnalogServer">
-		<param name="period">       10  				</param>
-		<param name="deviceId">     left_arm				</param>
-		
-		 <paramlist name="networks">
-		      <!-- elem name hereafter are custom names that live only in this file, they are used in the attach phase -->
-			<elem name="FirstSetOfJoints">  0  3  0  3 </elem> 
-			<elem name="SecondSetOfJoints"> 4  6  0  2 </elem> 	
-		    </paramlist>
-
-		<param name="channels">       7  				</param>
-		<param name="name"> 	     /left_arm				</param>
-
-		<action phase="startup" level="5" type="attach">
-		    <paramlist name="networks">
-			<!-- The param value must match the device name in the corresponding emsX file -->
-			 <elem name="FirstSetOfJoints">  left_upper_arm_mc </elem>	
-	     		 <elem name="SecondSetOfJoints"> left_lower_arm_mc </elem>  
-		    </paramlist>
-		</action>
-
-		<action phase="shutdown" level="5" type="detach" />
-	</device>
-
-
-
+    <device xmlns:xi="http://www.w3.org/2001/XInclude" name="left_arm_VFTserver" type="virtualAnalogServer">
+        <param name="period">       10  				</param>
+        <param name="deviceId">     left_arm				</param>
+        <paramlist name="networks">
+            <!-- elem name hereafter are custom names that live only in this file, they are used in the attach phase -->
+            <elem name="FirstSetOfJoints">(  0  3  0  3 )</elem>
+            <elem name="SecondSetOfJoints">( 4  6  0  2 )</elem>
+        </paramlist>
+        <param name="channels">       7  				</param>
+        <param name="name"> 	     /left_arm				</param>
+        <action phase="startup" level="5" type="attach">
+            <paramlist name="networks">
+                <!-- The param value must match the device name in the corresponding emsX file -->
+                <elem name="FirstSetOfJoints">  left_upper_arm_mc </elem>
+                <elem name="SecondSetOfJoints"> left_lower_arm_mc </elem>
+            </paramlist>
+        </action>
+        <action phase="shutdown" level="5" type="detach" />
+    </device>
+</devices>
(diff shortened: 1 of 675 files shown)


pattacini (approved):
Thanks heaps @triccyx 👍🏻 

This PR addresses https://github.com/robotology/robots-configuration/issues/222#issuecomment-744507821.

I would keep this PR open for a few days asking others to give it a glance 👀
Anyway, please don't merge the PR: I'll do that ⚠

cc @davidetome @S-Dafarra @violadelbono @julijenv

S-Dafarra:
Part of these modifications is already in ``devel``. Will this cause conflicts?

pattacini:
>Part of these modifications is already in devel. Will this cause conflicts?

Most likely yes @S-Dafarra, but I'll be dealing with those potential conflicts.

traversaro:
This PR is quite big, and seems to mix two things: 
* Re-formatting a lot of XML files according to some logic that it would be good to describe in the PR description (that is a good thing, to have consistency) 
* Address https://github.com/robotology/robots-configuration/issues/222#issuecomment-744507821 

Mixing the two is a bit dangerous as it is not possible to just check the changes related to the second point on their own. If this was done automatically, would it be feasible to separate the change in two commits, the first one that only changes the XML formatting, and the second one that just changes the `<elem name="FirstSetOfJoints"> 0  3  0  3 </elem>` to `<elem name="FirstSetOfJoints">( 0  3  0  3 )</elem>` ? If there is an easy command to reformat the XMLs, this would also help the future merge with devel, as we could first re-format also devel, and then see the comparison on the formatted XML. 

Furthermore, for keeping the consistency with the rest of the XML files and in general with the rest of files we have in all other repos, could it make sense to indent the XML tags with spaces rather then tabs?

pattacini:
Hi @traversaro 

@triccyx applied the fix using a script that resorts to xpath and a format change was expected.
We judged the implications carefully (we couldn't use grep) and then decided to go this way.

Don't know if what you asked is doable via xpath. @triccyx any comment on this?

traversaro:
> @triccyx applied the fix using a script that resorts to xpath and a format change was expected.

It would be nice to have a reference  to such a script, if available!

pattacini:
@triccyx perhaps you could replace tabs with a second pass using your module.

pattacini:
>It would be nice to have a reference to such a script, if available!

@triccyx churned out this Gitpod-enabled repo sporting the script/module.

traversaro:
Thanks, I think it is a good example of XPath use for processing XML files, can be used as a good model for other automatic modifications of XML files (as for example URDF or SDF files).

traversaro:
At a first glance, for just formatting the files without code modifications it should be sufficient to change this line to `changed = true` and comment out the following for: https://github.com/triccyx/fixname/blob/master/src/xmlfix.cpp#L23 .

For inserting spaces instead of tabs, probably it is sufficient to change the line in https://github.com/triccyx/fixname/blob/master/src/xmlfix.cpp#L46 to:
~~~
        bool res = root.save_file(toChange.c_str(), "    ");
~~~

triccyx:
> At a first glance, for just formatting the files without code modifications it should be sufficient to change this line to `changed = true` and comment out the following for: https://github.com/triccyx/fixname/blob/master/src/xmlfix.cpp#L23 .

Yes of course. In order to minimize the file changed I put false. But if we want to format all the files we can do this.
 
> For inserting spaces instead of tabs, probably it is sufficient to change the line in https://github.com/triccyx/fixname/blob/master/src/xmlfix.cpp#L46 to:
> 
> ```
>         bool res = root.save_file(toChange.c_str(), "    ");
> ```

Again yes.
We can abort this PR and redo with this second hint. The first hint should be done on all files, as a second step? 
Also do not forget the not well-formatted files.
In the case I suggest to extend the issue:
- format all files
- check for wrong files

pattacini:
I might have misunderstood before but I'm pretty much against modifying all the files. Let's stay tuned on the wrappers, please.

So:
- don't open up another PR: we can profitably update this PR. 
- we can replace tabs w/ spaces within the wrappers we touched.
- manually we can fix wrappers that have errors inside.

pattacini:
What @traversaro meant to ease the review was:
- sticking to wrappers only
- proposing a PR with 2 commits:
  - 1 commit devoted to formatting only
  - 1 commit devoted to the actual change

If it's too complicated to do that now, we can simply try to fix tabs as well as malformed XML files.

traversaro:
Thanks @pattacini, that is correct.

triccyx:
Perhaps the situation is not clear, it is better to have a discussion f2f (I will be present today at 15).
However:
without changing the script (the regexp part) it is not possible to reformat the 650 scripts removing the tab and put spaces.
As a matter of fact, if we put:
```
changed = true
```
all the other about 1000 XML will be modified (the find command return about 1600 XML).

If the request is:
```put the parenthesis and maintain space instead of tab```
The correct solution is to execute again the script (about 1 sec of work) on a clear git clone and redo PR.
Then we can commit and do the new formatting part issue if we want.

Otherway I need to change and test the script but this is an unuseful/dangerous work.

**Comment by traversaro** [comment]
> As a matter of fact, if we put:
> ```
> changed = true
> ```
> all the other about 1000 XML will be modified (the find command return about 1600 XML).

Yes, that was just a suggestion based on a first glance I gave at the code, but I was wrong. The change that I wanted instead to suggest to just format all the files that would be touched by this PR is to comment out the line `           current.node().first_child().set_value(out.c_str());` so that the only change will be the re-formatting, commit the formatting changes, and then re-run the script after uncommenting `current.node().first_child().set_value(out.c_str());` , and commit the resulting change. In this way, you would have two commits: 
* One commit only changing the syntax, but not for all the files, just for the one that is necessary to change
* A second commit just fixing the parenthesis, that would be much easier to review on its own.

However if this turns out to be to difficult/dangerous, feel free to proceed as it is, no problem.

**Comment by triccyx** [comment]
If you agree @traversaro, I prefer to reset the commit and apply the script again with space instead of tab as @pattacini suggest.
Where script==(c++ fixapp)+(bash find)

**Comment by traversaro** [comment]
> reset the commit and apply the script again with space instead of tab as @pattacini suggest.
> Where script==(c++ fixapp)+(bash find)

Ok, whatever it is easier for you. For the future, however, try to remember that is always a good idea to separate stylistic changes from actual changes, as this drastically simplifies the review process.

**Comment by triccyx** [comment]
Following @traversaro suggestion, for tab and space, I reset my branch and do again the modification.
```
git checkout fix-app
git reset --hard HEAD~1
git push -f origin fix-app
<do again modifications>
git commit -am "fix parentheses and keep space"
git push origin fix-app
```

triccyx:
Last 18 wrong files fixed and parentheses added

pattacini:
Thanks @triccyx 👍🏻 

Inspecting all the files still remains difficult but I've cherry-picked some to check the differences and we're ok ✔

Merging.

@S-Dafarra I'll deal with `devel` soon afterward.

PR #229, opened 2021-01-12
Added configuration of upperarms skin patches 

simeonedussoni:
I added the configuration for the upperarm skin patches.
- edited network configuration
- added starting values of parameters (they may be optimized by the user since we are not using the skin fabric)
- configured setup for front cover only
- modular approach: only upperarm skin patches are inserted, no modification on electronics/hardware side files apart from inserting the MTB4 in the list of the connected boards.


simeonedussoni:
I can not add revuiewers or assign the PR cc @pattacini

pattacini (changes requested):
thanks @simeonedussoni 👍🏻 

Apparently, you have removed files like `nao_*` from the robot `iiwaBoston01`.
Could you reinstate them, please?

simeonedussoni:
> 
> 
> thanks @simeonedussoni 👍🏻
> 
> Apparently, you have removed files like `nao_*` from the robot `iiwaBoston01`.
> Could you reinstate them, please?
> 
> (just update your fork and the PR will get automatically updated too)

Actually it's not needed there and is only a duplicate of the NAO conf files, if it's just to keep separarted work on iCub3 from other stuff I can reinstate them
BTW looking at the hardware side side we found an error in the cabling scheme that affects the configuration files, I also correct this.

pattacini:
>Actually it's not needed there and is only a duplicate of the NAO conf files, if it's just to keep separarted work on iCub3 from other stuff I can reinstate them

Ok leave them in then!

pattacini (approved):

PR #230, opened 2021-01-18
iCubGenova09 corrected parameters in skin configuration files

simeonedussoni:
After testing on the real robot I updated a few worng parameters escaped in the original work, namely:
1. the CAN bus is consistently set to `2 ` on all files
2. the CAN ID is fixed in every file
3. I put as default the configuration with all triangles activated even though only the ones for the front patch are currently present (there is no concern about bus throughput since there is only a couple of patches)


simeonedussoni:
memento: these are modifications implemented locally on icub-head machine and duplicated here to have a cleaner history.

pattacini (approved):
Thanks @simeonedussoni 👍🏻 

Actually, this is the mainstream repo so the information contained herein rules over everything else (there's no duplication). Thus, any downstream forks will need to be updated accordingly (cc @S-Dafarra).

PR #239, opened 2021-02-15
Changes in skinGui configs for naoPrague01, added positions of taxels

rustlluk:
Updated skinGui .ini files for naoPrague01. Now the GUI is correctly rotated and scaled.
Also positions of taxels and skinManaget .ini file for this robot are added to keep them in on place with other files for this robot.


pattacini (approved):
Thanks for the fix!

PR #240, opened 2021-02-22
Fix/camera config erzelli02

Nicogene:
The `grabber` wrapper has been deprecated in yarp 3.3 (see https://github.com/robotology/yarp/commit/1ad59e91f88b642f617f143b0493531cac8d9b7d)

Sorry for the multiple commits, they can be squashed in one.

### DIFF
--- a/iCubErzelli02/camera/dragonfly2_config_left.ini
+++ b/iCubErzelli02/camera/dragonfly2_config_left.ini
@@ -1,4 +1,4 @@
-device grabber
+device grabberDual
 subdevice dragonfly2
 width 320
 height 240
--- a/iCubErzelli02/camera/dragonfly2_config_left_bayer_320_240.ini
+++ b/iCubErzelli02/camera/dragonfly2_config_left_bayer_320_240.ini
@@ -1,4 +1,4 @@
-device grabber
+device grabberDual
 subdevice dragonfly2raw
 width 320
 height 240
--- a/iCubErzelli02/camera/dragonfly2_config_left_bayer_640_480.ini
+++ b/iCubErzelli02/camera/dragonfly2_config_left_bayer_640_480.ini
@@ -1,4 +1,4 @@
-device grabber
+device grabberDual
 subdevice dragonfly2
 width 640
 height 480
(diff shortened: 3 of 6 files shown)


pattacini (approved):

traversaro:
There is any specific reason for which on iCubErzelli02 the `dragonfly_***.ini` files spawn a network server wrapper YARP device(i.e. `grabber` or `grabberDual`) while on other robots such as iCubGenova02 the `dragonfy_***.ini` file only spawns the physical YARP device `dragonfly2` (see for example https://github.com/robotology/robots-configuration/blob/master/iCubGenova04/camera/dragonfly2_config_left.ini)?

traversaro:
> 
> 
> There is any specific reason for which on iCubErzelli02 the `dragonfly_***.ini` files spawn a network server wrapper YARP device(i.e. `grabber` or `grabberDual`) while on other robots such as iCubGenova02 the `dragonfy_***.ini` file only spawns the physical YARP device `dragonfly2` (see for example https://github.com/robotology/robots-configuration/blob/master/iCubGenova04/camera/dragonfly2_config_left.ini)?

Related files: 
* https://github.com/robotology/icub-main/blob/4215771441724b7ece6412d56d7f884f70cca703/app/default/scripts/cameras_calib.xml.template
* https://github.com/robotology/icub-main/tree/4215771441724b7ece6412d56d7f884f70cca703/app/default/scripts 

Probably it could make sense to open a separate issue to track this?

pattacini:
I think this PR only fixes the `grabber` in `grabberDual` but the correct usage appears different as @traversaro mentioned.

The device `grabberDual` should go into the file `ServerGrabberDualDragon.ini` instead, and it is already there.
That said, the option `subdevice` should be turned into `device` in files `dragonfly2_***.ini`.

cc @Nicogene

Nicogene:
I agree, the correct usage should be run the `yarpdev` with the `ServerGrabber*.ini` that will take the dragonfly2_***.ini, unfortunately, the yarpmanager XML files of `iCubErzelli02` are quite old(e.g `iCubStartup`, `CalibCameras` etc.) and they call directly
 dragonfly2_***.ini.

In any case this
> That said, the option subdevice should be turned into device in files dragonfly2_***.ini.

should be ok because the default wrapper is `grabberDual` https://github.com/robotology/icub-main/blob/122e11a3147d406e636020a377405086b95197b2/src/libraries/icubmod/dragonfly2/CMakeLists.txt#L20

Issue #247, opened 2021-03-24
Request for Comments: Clarify location of robot-specific manually added files?


traversaro:
Historically, the robots-configuration has been used to host two kind of files:
* The "official" (with a lot of quotes : ) ) configuration files for a given robot, that are typically maintained by the iCub Tech, and that routinely need to be massively updated for several reason (bugs found, change in the software that requires update in the configuration files). This files are not exactly the same for two different (even if identical) robots, due to calibration settings and similar robot specific parameters, but at least their structure is (or should be) consistent  across robot of the same version. 
* Robot-specific configuration and material: this includes robot-specific material that is maintained by specific Research Lines that used the robot (in IIT) or robot groups. While one school of thought is that this material should be in repositories that are maintained by the groups that are responsible for them, in practice for configuration files that are part of the "robot setup" it is quite convenient to be able to managed them with the rest of the robot configuration files.

This is de-facto situation. However, I think that a possible improvement (especially to simplify the possibility of doing **automatic changes** to the "official" configuration files) is to **dedicate a given directory in the robot directory for robot-specific files**, something called `extra` or similar ([bikeshedding](https://www.youtube.com/watch?v=5ouopR_eG08) on the name is welcome), in which all the robot-specific files can be placed. In this way, it would be easy to separate between "official" iCub configuration files and robot-specific ones.

traversaro:
Comments on this proposal are welcome: @Nicogene @pattacini @marcoaccame @S-Dafarra @violadelbono @davidetome @triccyx (feel free to tag more people that could be interesting).

pattacini:
Good suggestion @traversaro 👍🏻 
Also, `extra` sounds good.

traversaro:
>  Also, `extra` sounds good.

I must confessed that I borrowed the concept from the FMI3 standard, see [Section 2.5.1.6. Directory Extra](https://fmi-standard.org/docs/3.0-dev/#structure-of-zip).

S-Dafarra:
Sounds good. Among the files that could go in that directory there would be also the ``yarpmanager`` application files for example (since those are usually modified to consider the specific machines in the network for example). Hence this repo would store the configuration files in a broader sense, not just related to the robot itself, but to the entire setup.

Nicogene:
I like the idea, the `extra` folder could contain all the "extern" configuration files stored with the contexts hierarchy

pattacini:
@Nicogene if you mean the .ini files, that's not the case!
[...]

PR #251, opened 2021-04-07
Add extra folder to iCubGenova04

S-Dafarra:
This goes in the direction of https://github.com/robotology/robots-configuration/issues/247. It adds a folder called ``extra`` with a series of scripts that should help the use of the robot. It also contains the cluster-config and the yarpamanger application that are used in the setup.


pattacini (approved):

lrapetti:
Some of the scripts added with this PR are general enough to be used, as they are, with other robots (e.g. I have used `checkConfigurationFiles.sh` with `iCubGenova02`).

@pattacini @S-Dafarra do you think it makes sense to replicate them in other robots configuration folder, or could we move them out of the specific robot folder?
(also let me know if you want to move the discussion in a more visible place)

pattacini:
>@pattacini @S-Dafarra do you think it makes sense to replicate them in other robots configuration folder, or could we move them out of the specific robot folder?
(also let me know if you want to move the discussion in a more visible place)

For the time being, just replicate them. We might come up with a different structure in the future.
As of now, each robot has its own structure.

S-Dafarra:
> For the time being, just replicate them. 

I agree! When I wrote the files, I tried to be as generic as possible, but maintaining them for all the robots might be tricky as the configuration may be different (use of the superbuild, installation in local, setup specific information like the joystick address, ...). For the time being I think that a copy/paste would be fine.

PR #257, opened 2021-04-30
Update iCubGenova02 legs low-level torque control gain

lrapetti:
With this PR I update the torque-control gains for the `iCubGenova02` robot.

Not sure if I should be addressing `master` or `devel`


pattacini (approved):
`master` is ok 👍🏻 
I'll be porting these mods into `devel` as well.

PR #271, opened 2021-06-08
Sign change for torso pitch joint on all R1 robots

randaz81:
This PR has two main purposes:

1) swap the front and back lidars on R1 SN003 (they were not properly connected)

2) fix a **MAJOR** issue on the sign of the torso pitch joint (affecting all R1 robots) which was not correct according to the kinematic definition of the robot. Please refer to https://github.com/robotology/cer/issues/121 for further details. 
Please note that this change may affect the behavior of **user modules** that control the R1 pitch torso joint!


pattacini (approved):
Ok for me 👍🏻 

>Please note that this change may affect the behavior of user modules that control the R1 pitch torso joint!

This might be a bit critical as for example reaching module may be affected as you noticed.
I don't have time now to check this, unfortunately.

pattacini:
I'll have a look at it soon.

PR #272, opened 2021-06-14
[Experimental setup wristmk2_handmk3] fixed the name of the joints in the hand mk3

marcoaccame:
Aligned the names of the joints of the hand mk3 to the chosen convention.
In particular we control the movement of the following joints:

image

| Ref. | Joint name                   |
| ---- | ---------------------------- |
|      |                              |
|      |                              |
| P1   | l_hand_pinky_proximal_joint  |
| M1   | l_hand_medium_proximal_joint |
| i1   | l_hand_index_proximal_joint  |
| i3   | l_hand_index_adduction_joint |
| T1   | l_hand_thumb_proximal_joint  |
| T3   | l_hand_thumb_rotation_joint  |
| X1   | l_hand_metacarpus_joint      |

### DIFF
--- a/experimentalSetups/wristmk2_handmk3/hardware/mechanicals/hv3_left_hand-mcp9-mec.xml
+++ b/experimentalSetups/wristmk2_handmk3/hardware/mechanicals/hv3_left_hand-mcp9-mec.xml
@@ -6,18 +6,18 @@
         <param name="MotioncontrolVersion"> 6 </param>
         <param name="Joints"> 4 </param> 
         
-        <!-- joint number in sub-part       0             1               2               3             -->
-        <!-- joint name                                                                                 -->         
-        <param name="AxisMap">              0               1               2               3           </param>
-        <param name="AxisName">             "l_thumb"       "l_index"       "l_medium"      "l_pinky"   </param>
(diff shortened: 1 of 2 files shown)


traversaro:
I may be wrong, but the table in the docs seems to refer to physical joints of the urdf model, while the YARP interface (if I am not wrong) should refer to coupled axis), and I do not know if there is a table describing the relation between the physical joints and the controlled axis?

pattacini:
>I may be wrong, but the table in the docs seems to refer to physical joints of the urdf model, while the YARP interface (if I am not wrong) should refer to coupled axis), and I do not know if there is a table describing the relation between the physical joints and the controlled axis?

@marcoaccame @ale-git any comments on this observation?

marcoaccame:
@traversaro: we actually control w/ `yarprobotinterface` only the physical joints I have written in the table [above](https://github.com/robotology/robots-configuration/pull/272#issue-669435374). The other joints in the picture are all dependent from those according to mechanical coupling.

traversaro:
> @traversaro: we actually control w/ `yarprobotinterface` only the physical joints I have written in the table [above](https://github.com/robotology/robots-configuration/pull/272#issue-669435374). The other joints in the picture are all dependent from those according to mechanical coupling.

Ok, thanks!

traversaro (approved):

PR #291, opened 2021-09-15
iCubNancy01 - final configuration files

Uboldi80:
New configuration files for iCubNancy01

the Matlab live script report and relate files will be provide to @pattacini in order to generated a new tag.

cc: @davidetome


pattacini (approved):
Thanks @Uboldi80 @davidetome 
The new tag is upcoming.

PR #294, opened 2021-09-20
ROS robot_state_publisher deprecation fix

DatSpace:
Fixes error on any ROS above Melodic that doesn't recognize the robot state publisher type: state_publisher.

This was apparently an alias that has been deprecated for years and was removed with ROS Noetic. The robots still use Melodic so this was not a problem yet. It should not be a problem though. Tested in simulation.

https://github.com/ros/robot_state_publisher/pull/87  --> Look at number 5.


pattacini (approved):
Ok for me!

pattacini:
cc @randaz81

PR #297, opened 2021-10-13
iCubGenova01_nws

randaz81:
wip

PR #298, opened 2021-10-13
iCubGenova01 Add xls file containing formulas for raw encoder calibration

xEnVrE:
As per the title.

cc @Uboldi80 @randaz81


pattacini:
To be integrated also w/ #297.

PR #307, opened 2021-11-12
Hard stop calibration configuration for iCubGenova09 legs.

ale-git:
The iCubGenova09 legs are now calibrated with type 10 (hard stop) as in R1 pronosupination.

PR #308, opened 2021-11-12
iCubGenova09 updates after Venice video

S-Dafarra:


pattacini:
Hi @S-Dafarra 

There are a bunch of conflicts.
Could you please have a look at them?

pattacini (changes requested):
Hi @S-Dafarra 

There are a bunch of conflicts.
Could you please have a look at them?

S-Dafarra:
Hi @pattacini, yes I noticed. I thought of opening this PR to tackle these conflicts in view of https://github.com/robotology/robots-configuration/pull/307. Actually, I would need the @ale-git help on this. The version of this branch is the one used currently on the robot.

I know we edited the ``calibrationDelta``, but all the other conflicts are a bit obscure to me. If you could provide some insights that would be great

ale-git:
Maybe I can compile a little "how-to" change the calibration mode from absolute (type 12) and hard-stop (type 10), so that people can understand what's going on?

pattacini:
> Maybe I can compile a little "how-to" change the calibration mode from absolute (type 12) to hard-stop (type 10), so that people can understand what's going on?

That would be fantastic indeed 👍🏻

pattacini:
Hi @S-Dafarra 

>I know we edited the calibrationDelta, but all the other conflicts are a bit obscure to me. If you could provide some insights that would be great

Most likely, the conflicts are generated by #307 that has been merged.

My proposal is:
- if you'd be willing, you could propose a PR containing just those files that have been not touched by #307 (if there are any).
- for the remaining file, either we can really on the upcoming guide, or we can help the next week during the delivery.

S-Dafarra:
> Maybe I can compile a little "how-to" change the calibration mode from absolute (type 12) to hard-stop (type 10), so that people can understand what's going on?

That would help, but I also see some unexpected changes, like:
- the hip roll lower limit changed from -15 to -5 (quite critical for the walking)
- the hip pitch upper limit changed from 110 to 90
- the ``Gearbox_E2J`` changed from 1 to 64 (does this affect the computation of the joint velocity?)
- the resolution and the tolerance of the encoders changed

ale-git:
Hi @S-Dafarra:
- the -5 deg hip roll limit is the value in robotology, feel free to change it (the physical limit is about -20 deg)
- the +90 deg hip pitch is the physical limit, there is no way you can go higher :)
- the Gearbox_E2J is correct, it means that the 14 bit resolution of the AMO is repeated 64 or 32 times per physical revolution. The velocity isn't affected since this information is used at lower level to compute a 16 bit absolute angle measure.
- the resolution (16384) reflects the 14 bit per sector of the AMO, that multiplied by 64 sectors gives the same 2^20 former resolution.

S-Dafarra:
Ok, great, thanks @ale-git. Then I think I can resolve the conflicts myself!

PR #310, opened 2021-11-19
iCubGenova01 - Update right thumb configuration

xEnVrE:
After fixing https://github.com/robotology/icub-tech-support/issues/1236,

this PR update the Calibration for the right thumb proximal and distal.

It also updates the PID gains and the `ampsToSensor` parameter fof the right thumb opposition and proximal.


xEnVrE:
**Important**
These changes should be ported inside the new iCubGenova01nws folder as well (see #297)

pattacini (approved):

PR #311, opened 2021-12-10
Restored missing "ground_link" to cer.urdf files

elandini84:
I just restored a previously present link that has been removed from almost all the R1 `.urdf` files.

### DIFF
--- a/R1SN000/cer.urdf
+++ b/R1SN000/cer.urdf
@@ -29,6 +29,13 @@
     <child link="base_link"/>
     <dynamics damping="0.1"/>
   </joint>
+  <link name="ground_link"/>
+  <joint name="ground_link_fixed_joint" type="fixed">
+    <origin xyz="0 0 -0.2" rpy="0 0 0"/>
+    <parent link="base_link"/>
+    <child link="ground_link"/>
+    <dynamics damping="0.1"/>
+  </joint>
   <link name="mobile_base_lidar"/>
   <joint name="mobile_base_lidar_fixed_joint" type="fixed">
     <origin xyz="0.07 0 0.031" rpy="0 0 0"/>
--- a/R1SN002/cer.urdf
+++ b/R1SN002/cer.urdf
@@ -29,6 +29,13 @@
     <child link="base_link"/>
     <dynamics damping="0.1"/>
   </joint>
+  <link name="ground_link"/>
+  <joint name="ground_link_fixed_joint" type="fixed">
+    <origin xyz="0 0 -0.2" rpy="0 0 0"/>
+    <parent link="base_link"/>
+    <child link="ground_link"/>
+    <dynamics damping="0.1"/>
+  </joint>
   <link name="mobile_base_lidar_F"/>
   <joint name="mobile_base_lidar_F_fixed_joint" type="fixed">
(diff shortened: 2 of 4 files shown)


pattacini (commented):
Hi @elandini84 

Thanks for the PR.

I've got a couple of questions:
- You didn't extend the addition of the link to SN001, apparently. Is there any specific reason for that? Otherwise, we may complete the job.
- Was there any reason as to why the link got removed in the past? Calling also for other users' comments here: cc @randaz81 @vvasco.

elandini84:
> Hi @elandini84
> 
> Thanks for the PR.
> 
> I've got a couple of questions:
> 
> * You didn't extend the addition of the link to SN001, apparently. Is there any specific reason for that? Otherwise, we may complete the job.
> * Was there any reason as to why the link got removed in the past? Calling also for other users' comments here: cc @randaz81 @vvasco.

Hi @pattacini. I did not modify the SN001 file because the link and the joint are already there for that robot.

vvasco:
Hi @elandini84 @pattacini @randaz81. 
I'm not sure if there is a relation between these `urdf` models and those used in [`cer-sim` ](https://github.com/robotology/cer-sim/tree/master/gazebo)repo. 
However, in the `cer-sim` models the link to the ground would prevent the robot from navigating, probably this is why we don't have it there.
@elandini84  is there any specific reason why you need the link to the ground?

elandini84:
> Hi @elandini84 @pattacini @randaz81. I'm not sure if there is a relation between these `urdf` models and those used in [`cer-sim` ](https://github.com/robotology/cer-sim/tree/master/gazebo)repo. However, in the `cer-sim` models the link to the ground would prevent the robot from navigating, probably this is why we don't have it there. @elandini84 is there any specific reason why you need the link to the ground?

Hi @vvasco,
I need the `ground_link` for a navigation related application. I use the transformation from the `depth_center` frame to the `ground_link` one since I use point cloud data obtained from the robot RealSense depth images to check the "floor height" in order be able to correctly label areas where the robot can safely navigate. I don't know if the explanation is clear enough...

pattacini:
So, I would suppose that the `ground_link` is moving jointly with the physical robot instead of anchoring it to the ground preventing navigation as it seems to happen with the simulated robot in `cer-sim` as @vvasco mentioned.

pattacini:
> So, I would suppose that the `ground_link` is moving jointly with the physical robot instead of anchoring it to the ground preventing navigation as it seems to happen with the simulated robot in `cer-sim` as @vvasco mentioned.

Well, actually, if you aim to mark areas irrespective of the robot location, the `ground_link` is anchored to the ground and doesn't move jointly, maybe.

pattacini:
At any rate, if we're aligned and it's clear to us that the `ground_link` has different purposes and effects with the physical and the simulated robot, I'll be happy to merge.

elandini84:
> At any rate, if we're aligned and it's clear to us that the `ground_link` has different purposes and effects with the physical and the simulated robot, I'll be happy to merge.

The ground link is connected to the robot base_link through a fixed joint, so it moves with the robot.

pattacini (approved):

vvasco:
Thanks @elandini84, that's ok for me

PR #313, opened 2021-12-10
iCubGenova01 - Update right pinky configuration

xEnVrE:
As per the title, after fixing https://github.com/robotology/icub-tech-support/issues/1253.

cc @gsisinna


xEnVrE:
**Important**

These changes should be ported inside the new iCubGenova01nws folder as well (see #297)

pattacini (approved):

PR #315, opened 2021-12-13
Merge devel_iCubGenova04

gabrielenava:
I open this PR after the iRonCub experimental campaign to align `devel` with the `ami-iit` branch which is used on iCubGenova04.

cc @S-Dafarra


gabrielenava:
cc @HosameldinMohamed

PR #325, opened 2022-01-24
Add config files for wrist-setup: ems + amc_bldc

sgiraz:
These configuration files are intended to use with the `wrist-setup` shown during the presentation of `PI-09`.

Actually, the configuration allows to move a single joint of the wrist, but soon they will be changed in order to move the whole wrist.


pattacini (changes requested):
Hi @sgiraz 

Thanks!

Would you mind changing `wristmk2_handmk3` into `wristmk2_handmk3_ems_2foc` in order to express the difference?

pattacini (approved):

PR #338, opened 2022-02-17
iCubErzelli03: fix configuration file for the eyelids after the software/firmware upgrade

Nicogene:
These changes were necessary after the upgrade of the os to 20.04 and the SW/FW to the `v2021.11.1` distro release.

Please review code.


pattacini (approved):
Checked against iCubGenova09 👍🏻

PR #344, opened 2022-03-03
[iCubErzelli02] - Fix right thumb coupling matrix

gsisinna:
After receiving a maintenance request for the right thumb of iCubErzelli02 I checked the file in `/hardware/mechanicals` regarding the coupling matrix for the thumb, and this configuration with the minus sign leads to correct functioning without oscillations.
 
Probably in the past the sign has been changed according to the winding of the cable on the motor pulley. The robot has been tested and all the joints calibrate normally.


pattacini (approved):
Good 👍🏻

PR #348, opened 2022-03-29
skinSpec mods done 

davidetome:
This PR modifies the skinSpec files for the arms

### DIFF
--- a/experimentalSetups/iCubGenova02-VelCtrl/hardware/skin/left_arm-eb2-j4_15-skinSpec.xml
+++ b/experimentalSetups/iCubGenova02-VelCtrl/hardware/skin/left_arm-eb2-j4_15-skinSpec.xml
@@ -13,8 +13,20 @@
         </group>
         
         
+	<group name="specialCfgBoards">           
+	    <!-- it tells some boards to be special. 
+		 we can configure only: tx period, skinType, noLoad value. 
+		 rule #1 -> CAN1:14 has: period = 50 ms, skinType = 1 (palm and fingertips skin), noLoad = 0xf0
+	      -->
+	    <param name="numOfSets"> 1 </param> 
+	    <!--                            patch       adr start   adr end     period      type    no_load     -->            
+	    <param name="boardSetCfg1">     1           14          14          50          1       0xf0        </param> 
+		                  
+	</group>
+	
+	
        <group name="defaultCfgTriangle">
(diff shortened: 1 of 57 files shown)


davidetome:
w/ the first commit I modified all the ETH robots.

Concerning the CAN robots, I noticed that the CAN ID is expressed in HEX format.

@maggia80 , I guess I have to respect this format also while adding the `<group name="specialCfgBoards">` group, am I right?

cc @pattacini @marcoaccame

marcoaccame:
> ```
> 0xE        
> ```

I would keep the IDs in decimal as usual

davidetome:
> > ```
> > 0xE        
> > ```
> 
> I would keep the IDs in decimal as usual

@marcoaccame thanks for replyimg, but probably I was not clear..  the `specialCfgTriangles` section present in the [CAN robots](https://github.com/robotology/robots-configuration/blob/e0cdbb3506c657cb4114101db4221407c73b0e7a/iCubAberystwyth01/hardware/skin/left_arm-cfw2_can7-skinSpec.xml#L25) contains the HEX format.
My doubt was if adding the `specialCfgBoards` to them I have to use the HEX format also in it instead of the DEC format I used for the ETH robots.

marcoaccame:
> > > ```
> > > 0xE        
> > > ```
> > 
> > 
> > I would keep the IDs in decimal as usual
> 
> @marcoaccame thanks for replyimg, but probably I was not clear.. the `specialCfgTriangles` section present in the [CAN robots](https://github.com/robotology/robots-configuration/blob/e0cdbb3506c657cb4114101db4221407c73b0e7a/iCubAberystwyth01/hardware/skin/left_arm-cfw2_can7-skinSpec.xml#L25) contains the HEX format. My doubt was if adding the `specialCfgBoards` to them I have to use the HEX format also in it instead of the DEC format I used for the ETH robots.

The parser of the xml is the same for all the robots, be them ETH or CAN. It supports the reading of both decimal and hex values in xml files. 

So the use of hex values or decimal is just a choice.

davidetome:
I've tested on the ETH robot in proto and it was ok but  on iCubGenova01(CAN) there were the patch ID not aligned to the added one.

All the CAN based robots needs to be fixed, and only few ETH needs to be fixed.

I'm gonna do it.

maggia80 (approved):

davidetome:
I've fixed w/ the last commit

PR #359, opened 2022-05-19
added robotinterface fix for odometry

ste93:
Fixed order of startup and shutdown of odometry in robotinterface


ste93:
@randaz81 @elandini84 @DatSpace fyi

pattacini (approved):

PR #385, opened 2022-09-21
Wrist mk2: Update position PID values 

mfussi66:
This PR updates the values of the position PID of the wrist mk2 for all three joints. The values were found by automatically tuning the controllers on an identified linear model.

The improvements are significant:

## Test on the wrist

wristtraj

wristerr

wristeul

## Metrics
I moved only the roll DoF, so ideally I would like the others to be zero.
To compare the results, I can compute the MSE with respect to a zero reference of the yaw and pitch DoFs:

| UoM: (deg^2) |Before| After|
| --- | --- | --- |
| **Yaw** |0.234| 0.0432|
| **Pitch** |0.126| 0.0502|


mfussi66:
cc @pattacini

pattacini (approved):

PR #386, opened 2022-09-21
iCubGenova09: add files for running the pylonCamera and some fixes for the calibrators for the new head

Nicogene:


pattacini (approved):
Thanks heaps @Nicogene 

cc @S-Dafarra

PR #390, opened 2022-09-29
iCubGenova09: remove the grabberDual ini files

Nicogene:
Since `grabberDual` gave us performance problems with the new cameras, and it is deprecated, I would remove the ini files.

cc @S-Dafarra @GiulioRomualdi


pattacini (approved):
Agreed 👍🏻

PR #395, opened 2022-11-11
new configuration files for all R1 robots

randaz81:
Include new `enabled_by` `disabled_by` attributes and various ROS2 nws


pattacini (approved):
Thanks heaps @randaz81 for these nice additions!

PR #398, opened 2022-11-14
ergocub: remove sign in calibration 12 and update pitch roll neck gains

mfussi66:
This PR fixes an important aspect in the calibration of the head joints of ergoCub, by removing the minus sign. This is because calibration type 12 needs the absolute value of the encoder.

Additionally, the PID values of the neck pitch and roll are updated to the latest autotuning results.


pattacini (approved):
Why still draft?

mfussi66:
> Why still draft?

Opened in draft accidentally, it's not wip.

PR #411, opened 2023-01-10
[iCubGenova11] Do not start legs by default

xEnVrE:
As per the title.

Given that we are using a new stand for the robot with variable height and that not all the legs configurations are now collision free, given the geometry of the stand, we would like to disable the legs by default.


xEnVrE:
cc @AntonioConsilvio

pattacini (approved):

PR #415, opened 2023-01-18
iCubGenova11: add mc and ft ros2 nws

Nicogene:
This PR adds the xml files for running the ft and mc ros2 nws.

It has been tested on the robot.

immagine

Probably more check/fixes are needed on the urdf in order to visualize the data correctly on `rviz`, meanwhile here is an example of visualization of data on `rqt`:

immagine

cc @maggia80 @pattacini @randaz81 @elandini84 @xEnVrE

### DIFF
--- a/iCubGenova11/icub_all_ros2.xml
+++ b/iCubGenova11/icub_all_ros2.xml
@@ -1,0 +1,146 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!DOCTYPE robot PUBLIC "-//YARP//DTD yarprobotinterface 3.0//EN" "http://www.yarp.it/DTD/yarprobotinterfaceV3.0.dtd">
+
+<robot name="iCubGenova11" portprefix="icub" build="1">
+    <params>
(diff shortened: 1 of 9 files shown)


randaz81:
regarding iCubGenova11/icub_all_ros2.xml I recommend using the syntax shown in https://github.com/robotology/robots-configuration/blob/master/R1SN003/CER.xml, so you can use a single main xml file.

pattacini (approved):
Super @Nicogene 🚀 

Regarding @randaz81's [suggestion][sugg], you can take it on now @Nicogene or we can, at a certain point, go through all robots for a general upgrade.

[sugg]: https://github.com/robotology/robots-configuration/pull/415#issuecomment-1386923482

Nicogene:
Yes I think we can takle that upgrade with the new sintax later on probably it needs also intensive testing

pattacini (approved):

randaz81:
I added the issue here: https://github.com/robotology/robots-configuration/issues/417

Issue #404, opened 2022-11-29
`controlboardwrapper2` deprecation


randaz81:
device `controlboardwrapper2` was deprecated in previous yarp releases and it will be completely removed in the upcoming yarp 3.8.
The correct devices to use are: `controlBoard_nws_yarp` / `controlboardremapper`.
An update of the robot configuration files is highly recommended, otherwise, the robot will be not able to startup anymore.

pattacini:
We ran a [campaign](https://github.com/robotology/robots-configuration/pulls?q=is%3Apr+is%3Aclosed+nws%2Fnwc) in the past to update the robots to the use of the NWS/NWC layers.

However, there are still some [leftovers](https://github.com/robotology/robots-configuration/pulls?q=is%3Apr+is%3Aclosed+nws%2Fnwc), which pertain to:

| Robots/Setups/Folders | Status | Notes |
| :--- | :---: | :--- |
| `iCubLisboa01` | ❌ | Needing fix |
| `R1SN000` | ❌ | Needing fix |
| `iCubTemplates` | ❌ | Needing fix |
| `iRobCub-Mk1` | ❔ | We may have kept it unchanged upon request |
| `experimentalSetups` | ❔ | We need to judge case by case as some setups are old |
| `vizzy` | 👍🏻 | We don't maintain it |

@davidetome who was the main developer in charge may add more insights.

Nicogene:
From [this](https://github.com/search?q=repo%3Arobotology%2Frobots-configuration+controlboardwrapper2&type=code) seems that the new `iCubGenova11` is using `controlboardwrapper2`.

Maybe the template for the new robots has to be updated? Otherwise we continue to introduce it

pattacini:
>Maybe the template for the new robots has to be updated? Otherwise we continue to introduce it

Very true!
I'll update the OP to include iCubGenov11 too.

cc @AntonioConsilvio @davidetome

davidetome:
> From [this](https://github.com/search?q=repo%3Arobotology%2Frobots-configuration+controlboardwrapper2&type=code) seems that the new `iCubGenova11` is using `controlboardwrapper2`.
> 
> Maybe the template for the new robots has to be updated? Otherwise we continue to introduce it

Ciao @Nicogene , actually the templates are not used to create a new robot folder, but only to explain files structure, params meaning and so on.
[...]

PR #416, opened 2023-01-18
Fix joint names for the hands

Nicogene:
The names have been fixed according to https://icub-tech-iit.github.io/documentation/icub_kinematics/icub-joints/icub-joints/#left-arm

These are the names also used in the urdf.

It is important that the template used for new robot is rebased also over these changes

cc @pattacini @AntonioConsilvio @xEnVrE


Nicogene:
(it seems that only `iCubGenova11` used wrong names)

pattacini:
>It is important that the template used for new robot is rebased also over these changes

I've put a note in #404's OP.

>(it seems that only iCubGenova11 used wrong names)

Apparently, iCubGenova11 was spawned from an old robot before we upgraded it.
See https://github.com/robotology/robots-configuration/issues/404#issuecomment-1381959622.

pattacini (approved):

PR #421, opened 2023-01-27
Ergocub fapreader

davidetome:
changes in XML files during fingers tuning made by @sgiraz

cc @pattacini @marcoaccame @AntonioConsilvio


pattacini (approved):
Changing the base branch from `devel` to `master` as this is a patch actually (and you probably started out branching from `master`).

Merging and deleting the compare branch.
I'll then update `devel` too.

Please, use forks (which can be shared among developers) 😃

PR #423, opened 2023-01-30
eCub 1.0 S/N:000 New torso calibration

AntonioConsilvio:
eCub 1.0 S/N:000 New torso calibration and new torso_head test file.


pattacini (approved):
Thanks @AntonioConsilvio 👍🏻 
Merging in `master` as this is more like a patch!

PR #431, opened 2023-02-02
ergoCub - aligning XML files

davidetome:
as per title

cc @pattacini


Nicogene (approved):
If it has been successfully tested on the robot, LGTM!

ergoCubSN000/hardware/inertials/head-inertial.xml
```diff
@@ -43,7 +43,7 @@
             </group>
 
             <group name="SETTINGS">
-                <param name="acquisitionRate">      10      </param>
+                <param name="acquisitionRate">      20      </param>
```
Nicogene, line 46: Just a curiosity, why the acquisition rate has been augmented to 20 (I suppose) ms?

sgiraz: Yep, it is interpreted as a period in milliseconds.
@marcoaccame and I applied this change to avoid possible issues when reading the IMU from the new RFE master.
We may restore it to 10ms, if necessary, once the IMU works correctly.

sgiraz (approved):
ok

ergoCubSN000/hardware/inertials/head-inertial.xml
pattacini: Uhm, I honestly avoid this. Let me revert this change. Local fiddling is always possible but changing the pristine original and intended value is risky.

pattacini, line 46:
```suggestion
                <param name="acquisitionRate">      10      </param>
```

pattacini (changes requested):
There are some conflicts I'm gonna deal with.

pattacini:
Hi @davidetome @mfussi66 

Eventually, could you please resolve the conflict on the file?

pattacini (changes requested):
The PR shows files that shouldn't belong to it.
Probably, the best thing is to reset your `fork/devel` onto the latest `upstream/devel` and then reapply the new changes.

PR #445, opened 2023-02-27
[ergoCubSN000] Low level joint position and wbd @500Hz

GiulioRomualdi:
This PR tunes the low-level position gains and calibrates the joint's position.

Moreover, in 34cd5b4ab13a716b49a5492e44a45dafc25f7826 we increased the low-level frequency (now WBD runs at 500Hz)

We tested the parameters with the walking-controller and we were able to run several successful experiments 

cc @lrapetti @mebbaid


pattacini (approved):
LGTM, although I would call others to review it before merging.

Nicogene (approved):

pattacini:
Merging anyway to speed up.
We can always refine later, if required.

PR #449, opened 2023-03-03
Full configuration file for test bench arm setup calib type6 FAP

MSECode:
This commit contains the complete files for testing calib type 6 FAP on the left arm test bench setup
Please keep connections and ports as specified in the files thus to do not have issues with connection errors
since boards have some issues on specific ports.
Update configuration to be aligned with new naming convention for axesNames and POS service


MSECode:
cc: @valegagge

pattacini (approved):

PR #455, opened 2023-03-10
Enable imu on mtb4

martinaxgloria:
As per the title and in https://github.com/icub-tech-iit/fix/issues/1339, the configuration files for enabling the IMU on the MTB4 boards are created.


pattacini:
Hi @martinaxgloria 

I've turned the PR into draft to indicate that we still need to perform testing.
Once done with the tests, you can turn it back to ready for review.

PR #459, opened 2023-03-20
ergoCubSN000: removed mtb4fap

marcoaccame:
This PR allows to use the `mtb4c` board in the hands of `ergoCub` rather than the `mtb4fap`, which is due to be removed from `robotology`.

The PR is bet merged after a run of `yarprobotinterface` and the reprogramming of the four `mtb4fap` boards in the two hands so that they become four `mtb4c` w/ the same CAN addresses. No other changes are required as long as the robot runs the latest `devel`.

### DIFF
--- a/ergoCubSN000/hardware/POS/left_hand-pos2.xml
+++ b/ergoCubSN000/hardware/POS/left_hand-pos2.xml
@@ -12,15 +12,15 @@
         <group name="PROPERTIES">
 
             <group name="CANBOARDS">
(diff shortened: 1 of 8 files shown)


pattacini:
I've put the PR in draft mode to indicate that we need to test it first!

sgiraz:
Hi @marcoaccame, I'm gonna test it tomorrow! Stay tuned, pls 🔈

PR #461, opened 2023-03-24
Use calib 6 for open-close finger joints right arm of ergoCubSN000

MSECode:
Adding calibration configuration files for right arm of ergoCubSN000 Setting calibration 6 for fingers adding parameters Setting rotor position limits min and max after first completed calibration Check correctness of calibration procedure on ergoCubSN000

### DIFF
--- a/ergoCubSN000/calibrators/left_arm-calib.xml
+++ b/ergoCubSN000/calibrators/left_arm-calib.xml
@@ -11,27 +11,27 @@
 	</group>
 
 	<group name="HOME">
-		<param name="positionHome">            5         30         0          10         0        0       0     30.00      30.00   30.00   30.00   30.00   30.00  </param>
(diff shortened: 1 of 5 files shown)


MSECode:
In left arm removed joint 9 (not 7 as in commented line) from calibration order because encoder not reading. (The problem should be on motor enc not on fap enc but should be re-checked on robot).
Anyway whole calibration procedure checked and working.

Add who should review as reviewer. I cannot add anyone (even if I should have  permits) 

cc: @pattacini @sgiraz @valegagge

pattacini (approved):

sgiraz:
> The problem should be on motor enc not on fap enc but should be re-checked on robot

Hi @MSECode,
On the left-hand everything is working correctly. The only thing to do is a re-calibration of the index-adduction. I can take care of that.

PR #467, opened 2023-03-30
[ergoCubSN000] Fix right arm calib sequence on branch devel

AntonioConsilvio:
### What changes?
During last [merging](https://github.com/robotology/robots-configuration/commit/3d1bdb2ce6eb8592f5d0b442874d50d1345fac21) of branch master into devel, two calib orders were created in the left arm calibration file in the devel branch.

This PR fix this problem.

cc @sgiraz


pattacini (approved):
My bad, thanks for the fix!

PR #478, opened 2023-04-14
Adding conf to arm bench setup for telemetry device dumper

MSECode:


MSECode:
add conf file for Telemetry Device Dumper to arm bench setup
@valegagge @pattacini

pattacini (approved):

PR #483, opened 2023-04-18
[ergoCubSN000] Torso: changing of current limits and more

AntonioConsilvio:
### What changes?
This PR change the current limit of the torso roll and the torso pitch motors. This PR also changes the gearbox ratio between the pitch motor and the joint, due to a hardware intervention.  

### Note
These new current limits are safe because they were extracted from the motor datasheets, but the pitch is not able to stay in position 0 for more then 4/5 seconds when the robot is in home position. It falls in HW fault for i2t.

cc @maggia80 @sgiraz @Fabrizio69

### DIFF
--- a/ergoCubSN000/hardware/motorControl/torso-eb5-j0_2-mc.xml
+++ b/ergoCubSN000/hardware/motorControl/torso-eb5-j0_2-mc.xml
@@ -16,10 +16,10 @@
         <param name="jntPosMax">                23                  45                  43                  </param>
         <param name="jntPosMin">                -23                 -18                 -43                 </param>
         <param name="jntVelMax">                120                 120                 120                 </param>
-        <param name="motorNominalCurrents">     5000                15000               5000                </param>
-        <param name="motorPeakCurrents">        10000               20000               10000                </param>
-        <param name="motorOverloadCurrents">    15000               30000               15000               </param>
-        <param name="motorPwmLimit">            10000               16000               10000               </param>
+        <param name="motorNominalCurrents">     3770                3770                5000                </param>
+        <param name="motorPeakCurrents">        14000               14000               10000                </param>
+        <param name="motorOverloadCurrents">    18000               18000               15000               </param>
+        <param name="motorPwmLimit">            10000               10000               10000               </param>
     </group>
 
     <group name="TIMEOUTS">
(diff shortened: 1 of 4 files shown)


ergoCubSN000/hardware/motorControl/torso-eb5-j0_2-mc.xml
```diff
@@ -16,10 +16,10 @@
...
         <param name="jntPosMin">                -23                 -18                 -43                 </param>
         <param name="jntVelMax">                120                 120                 120                 </param>
-        <param name="motorNominalCurrents">     5000                15000               5000                </param>
-        <param name="motorPeakCurrents">        10000               20000               10000                </param>
-        <param name="motorOverloadCurrents">    15000               30000               15000               </param>
-        <param name="motorPwmLimit">            10000               16000               10000               </param>
+        <param name="motorNominalCurrents">     3770                3770                5000                </param>
+        <param name="motorPeakCurrents">        14000               14000               10000                </param>
+        <param name="motorOverloadCurrents">    18000               18000               15000               </param>
+        <param name="motorPwmLimit">            10000               10000               10000               </param>
```
pattacini, line 22: Why did we reduce also the PWM limit?
By contrast, I would tune these value to the full scale for both the joints. Being controlled in current, there's no harm in doing so.

To be checked the full scale here.

pattacini (changes requested):

ergoCubSN000/hardware/motorControl/torso-eb5-j0_2-mc.xml
AntonioConsilvio: @pattacini We reduced the pwm limit to the same level as iCub3. In addition, the roll joint (which has the same motor as the pitch) has a power limit of 10000.  So we made this choice.

However, should we increase it?

maggia80: yes, I agree, we can keep it at full scale.

pattacini: Let's go to the PWM full scale then!
We need to confirm its value first.

pattacini: Let's set them up to `16000`. Even if it might not be the full-scale, it's certainly higher than `10000` and conservative wrt our old thresholds.

pattacini, line 22:
```suggestion
        <param name="motorPwmLimit">            16000               16000               16000               </param>
```

pattacini (approved):

PR #498, opened 2023-04-26
Make BUILD_TESTING OFF by default

traversaro:
By default `include(CTest)` makes `BUILD_TESTING` `ON`, but to avoid that users compile tests even if they do not use them, we can explicitly add the option and set it to `OFF`.


pattacini (approved):

pattacini:
Thank you!

PR #502, opened 2023-05-05
ergoCubSN000: modify alljoints-mc_remapper and torso-mc_remapper to not expose torso_pitch

traversaro:
These PR modifies the configuration files in `ergoCubSN000` to not expose the `torso_pitch` joint on YARP and ROS2 wrappers, while keep it in the configuration to minimize the amount of configuration files modified. 

To reach this goal, I migrated the `alljoints-mc_remapper` and `torso-mc_remapper` from specifying the joint that they control indirectly via the `networks` group, to explicitly specify the axes that they control via the `axesNames` parameter (see https://yarp.it/latest/classControlBoardRemapper.html#details for more details). For future mantainance, I also added commented the `axesNames` parameter in the case in which `torso_pitch` is actually present. In theory we could think of migrating all the other remappers from `networks` to `axesNames` for consistency, but this is out of the scope of this PR.

This PR ensures compatibility with the modifications in the simulation model done in https://github.com/icub-tech-iit/ergocub-software/pull/103 .

**I did not test this modification on the robot.**

PR #503, opened 2023-05-05
[ergoCubSN000] added missing joints param

pattacini:
This PR fixes #502 adding missing `joints` param.

The CI didn't catch it as it was running based on `pull_request_target` event (to fetch an env var). Previously fixed it via https://github.com/robotology/robots-configuration/commit/f8194810933e643d9a61bdbeb177a262d0aeb77f.


traversaro:
What was the error? If `jointAxes` is passed, in theory `controlboardremapper` does not need any `joints` parameter, see:
* https://www.yarp.it/latest/classControlBoardRemappe r.html#details
* https://github.com/robotology/yarp/blob/v3.8.0/src/devices/controlBoardRemapper/ControlBoardRemapper.cpp#L51
* https://github.com/robotology/yarp/blob/v3.8.0/src/devices/controlBoardRemapper/ControlBoardRemapper.cpp#L111

pattacini:
The error is exactly the missing `joints` param.

The XSD schema used to run the verification at the moment is not general enough to catch this peculiarity and it's also difficult to enhance it as we don't use XML as it ought to be used, unfortunately.

I saw that `joints` was already used in conjunction with `jointAxes`, although not required as you said. See for example:
- https://github.com/robotology/robots-configuration/blob/master/ergoCubSN000/wrappers/motorControl/left_arm-mc_remapper.xml

As of now, I need it to make the validation run flawlessly.

pattacini:
cc @Nicogene

traversaro:
> The error is exactly the missing `joints` param.
> 
> The XSD schema used to run the verification at the moment is not general enough to catch this peculiarity and it's also difficult to enhance it as we don't use XML as it ought to be used, unfortunately.
> 
> I saw that `joints` was already used in conjunction with `jointAxes`, although not required as you said. See for example:
> 
> https://github.com/robotology/robots-configuration/blob/f8194810933e643d9a61bdbeb177a262d0aeb77f/ergoCubSN000/wrappers/motorControl/left_arm-mc_remapper.xml#L5-L6
> 
> As of now, I need it to make the XML validation run flawlessly, even if it's not required.

Ah, ok, so the source of the check is https://github.com/robotology/robots-configuration/blob/f8194810933e643d9a61bdbeb177a262d0aeb77f/tests/dry-run/check-xml/schemas/remapper.xsd#L38 , I see.

traversaro:
A nitpick (that should not be a big problem as the value is actually ignored), probably it should be 45 as the original one (before removing the joint) was 46 (see https://github.com/robotology/robots-configuration/pull/502/files#diff-6dfa09278879a57fda38fbadef41b069ec9b0f5d0b9d111290eba6e1d2492813L13 )?

pattacini:
Counting them manually, it came out 44.

traversaro:
> Counting them manually, it came out 44.

Indeed, it seems that was wrong before, as the number of joints were 4+13+13+3+6+6 = 45 .

PR #510, opened 2023-05-12
ergocub: increase head pitch and roll position PID gains

mfussi66:
This PR proposes increased PID gains of the pitch and roll of the ergoCub head, for position control.

The gains were tested with the still robot during external shake-ups on the shoulders and the tilted hip pitch:
pidconfronto

And during the bend over sequence during the box lifting test:
piddevel
pidmagici


pattacini:
The test on walking was successful 🎉 
Let me turn this PR into ready for review.

We will keep testing the gains these days before merging.

pattacini:
The latest tests have confirmed that the new PID gains are working well.
cc @xEnVrE

Issue #509, opened 2023-05-11
CI failing on `ergoCubSN000` and `iCubGenova09` with the latest `yarp-3.8`


pattacini:
I've just regenerated the docker image used by the CI relying on the `yarp-3.8` branch rather than the `v3.8.0` release (#508) and I'm getting the following errors:

[dry-run_log_ergoCubSN000_dev.txt](https://github.com/robotology/robots-configuration/files/11454325/dry-run_log_ergoCubSN000_dev.txt):
```console
[ERROR] embObjMultipleFTsensors: BOARD left_arm-eb2-j0_1 (IP 10.0.1.2)   getTemperatureSensorMeasure() fails data for index: 0  not found
```

[dry-run_log_iCubGenova09_dev.txt](https://github.com/robotology/robots-configuration/files/11454356/dry-run_log_iCubGenova09_dev.txt):
```console
[ERROR] embObjMultipleFTsensors: BOARD left_leg-eb8-j3_5 (IP 10.0.1.8)   getTemperatureSensorMeasure() fails data for index: 0  not found
```

@traversaro, will https://github.com/robotology/yarp/pull/2967 fix this problem too?

traversaro:
This is more of an issue of `embObjMultipleFTsensors`, but indeed it will be mitigated by https://github.com/robotology/yarp/pull/2967 .

It is related to the second fix from https://github.com/robotology/icub-main/pull/871, and actually the reason why now it appears is that we fixed that bug. Before then, the value returned by `embObjMultipleFTsensors::getNrOfTemperatureSensors` was zero if no temperature measurement was received, and so no one called `embObjMultipleFTsensors::getTemperatureSensorMeasure` . https://github.com/robotology/yarp/pull/2967 ensures that `embObjMultipleFTsensors::getTemperatureSensorMeasure` is not called if the status of the sensors is not `MAS_OK`, so it should ensure that this error does not appear.

pattacini:
Thanks for the clarification!

Does your last doubt mean that the driver per se in this configuration is not ready to be dry run at all?

traversaro:
It is not the driver, it is the `multipleanalogsensorsserver` that will print errors if no valid sensor measure is available. I wonder how other network wrapper server deal with this, perhaps they do not complain and just do not publish data if it is not available?

pattacini:
Yep, with the term `driver` I was referring to the whole stack.

[...]

PR #512, opened 2023-05-15
ergoCub1.0 S/N:000 – Update and improve configuration files

AntonioConsilvio:
**What's new:**
- `head-calib.xml` for this [ issue](https://github.com/robotology/icub-tech-support/issues/1550) 
- `right_arm-eb22-j7_10-mec.xml` and "**right_arm-eb22-j7_10-mc_service.xml**" for this [issue](https://github.com/robotology/icub-tech-support/issues/1556) 
- `left_arm-eb25-j11_12-mc_service.xml` for this [issue](https://github.com/robotology/icub-tech-support/issues/1555) 
- the `2FOC` firmware version is now aligned in the configuration files with the one actually installed on the robot.

In addition, I added the original calibration order in a commented line for each calibration file; to avoid the risk of entering an incorrect calibration sequence in the event of temporary changes to the calibration order.

**Notes:**
-  ⚠️ I **deleted** many "custom" files as they were not maintained anymore.
This was done with the aim of reducing the number of files in the `ergoCubSN000` configuration folder to make navigation easier and smoother. Please check among the deleted files whether they are still in use, as in that case they can be restored.

cc @sgiraz @S-Dafarra @lrapetti @GiulioRomualdi @xEnVrE @mfussi66


ergoCubSN000/realsense.xml
pattacini, line 1: This name doesn't sound ok since it regards both lidar and realsense.
Can you clarify?

ergoCubSN000/calibrators/right_arm-calib.xml
```diff
@@ -31,7 +31,7 @@
 		<param name="startupPosThreshold">     2                 2              2           2          2             2              2            90           5           90           5           5           5           </param>
 	</group>
 
-<!--	<param name="CALIB_ORDER"> (4 5 6) (3) (2) (0) (1) (7 8) (9) (10 11 12) </param> --> 
+<!--	<param name="CALIB_ORDER"> (4 5 6) (3) (2) (0) (1) (7 8) (9) (10 11 12) </param> --> <!-- Don't remove this line -->
 	<param name="CALIB_ORDER"> (4 5 6) (3) (2) (0) (1) (7 8) (10 11 12) </param>
```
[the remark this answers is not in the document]
sgiraz: It should be the `index-adduction` that currently is not working due to https://github.com/robotology/icub-tech-support/issues/1522. We're already working on it in our internal activities.

ergoCubSN000/realsense.xml
AntonioConsilvio: Hi @pattacini, I have restored `realsense2.xml` and restored the name `sensors.xml` instead of `realsense.xml`.

ergoCubSN000/head.xml
```diff
@@ -19,10 +19,6 @@
     <xi:include href="wrappers/inertials/head-inertials_wrapper.xml" />
     <xi:include href="hardware/inertials/head-inertial.xml" />
 
-   <!-- Head sensors-->
-    <xi:include href="sensors/realsense.xml" />
-    <xi:include href="sensors/rplidar.xml" />
-
```
pattacini, line 25: Hi @AntonioConsilvio 

Could you clarify the rationale for this removal?
Thanks!

pattacini:
> > I think that the `realsense2.xml` file is the one currently used for running the realsense.
> 
> Indeed, it is also the one referred to in the ergoCub laptop "basic" xml loaded in the `yarpmanager` for running the RealSense on the ergocub-head.

@AntonioConsilvio reinstated the file.

ergoCubSN000/head.xml
AntonioConsilvio: Yes @pattacini . This `head.xml` file activates the neck motors and realsense. 

But it was unusable on both the Xavier and the torso CPUs, because if you start a `yarprobotinterface --config head.xml` on the torso CPU, it cannot start because the realsense is connected to the Xavier. 

At the same time, it cannot start on the Xavier because the motor control boards are connected to the torso CPU.

So this file was unusable.

I therefore removed the `head sensors` part so that the file could be started on the torso, and the realsense can be started separately on the Xavier with `realsense2.xml` or `sensors.xml`.

cc @mfussi66

pattacini: Thanks for the explanation!

S-Dafarra (approved):
It looks good to me, but I would also wait for the other reviewers as they use the robot more often

Nicogene (approved):
In light of @AntonioConsilvio / @sgiraz clarifications lgtm now!

sgiraz (approved):
Ok for me!

pattacini (approved):
Fine with me although I'm trying to reinstate the CI on ergoCubSN000, which I temporarily disabled earlier on (see #509).

Also, awaiting @xEnVrE's review.

pattacini:
Let me merge this PR straight away.
The CI will be reinstated soon.

PR #513, opened 2023-05-15
Add streaming of ergoCub leg imus

isorrentino:
As per the title. In addition, I aligned the name of the port to expose the right arm ft and the feet imu.

PR #515, opened 2023-05-16
ergoCubSN000: Fix torso calibrator

traversaro:
The PR https://github.com/robotology/robots-configuration/pull/502 removed `torso_pitch` from the `torso-mc_remapper` remapper device. However, the calibrator still expects three joints, so the best solution is to attach the calibrator directly to the hardware device instead of the remapper.

This should fix the problem reported in https://github.com/robotology/robots-configuration/pull/513#issuecomment-1549185646 by @isorrentino .


pattacini:
cc @isorrentino @AntonioConsilvio 

See https://github.com/robotology/robots-configuration/pull/513#issuecomment-1549185646.

pattacini (approved):

PR #519, opened 2023-05-17
[ergoCubSN000] Update the neck PID gains

GiulioRomualdi:
This PR updates the ergoCubSN000 neck gains with the one provided by @mfussi66


GiulioRomualdi:
We tested the gains while the robot was walking

pattacini:
Hi @GiulioRomualdi 

I've just merged #510, which applied the same update.

PR #523, opened 2023-05-19
Replace embObjStrain w/ embObjFTSensor

Nicogene:
Now embObjFTSensor can handle also strain board, see https://github.com/robotology/icub-main/pull/874

I will test it on a robot that has both strain and strain2 (e.g `iCubGenova02/4).


pattacini:
Hi @Nicogene 

I suppose that the CI will fail until https://github.com/robotology/icub-main/pull/874 is merged.
Correct?

Nicogene:
> Hi @Nicogene
> 
> I suppose that the CI will fail until [robotology/icub-main#874](https://github.com/robotology/icub-main/pull/874) is merged. Correct?

Yes it should 🤞🏻

valegagge:
Hi @Nicogene,
I have only a doubt: I saw you updated all the templates file replacing embObjStrain w/ embObjFTSensor. So here is my question: do we want to replace also the old templates?

In my mind, the old templates are related to old versions, so we don't need to update them, just the current template.

What do you think?

traversaro:
> In my mind, the old templates are related to old versions, so we don't need to update them, just the current template.

I think the question is: they are related to old versions in the sense of "old robots", or "old releases of the software" (O2)? 

If they are "old robots", perhaps it make sense to update them, if they are "old releases of the software", perhaps it could make sense to delete them from the repo, and if people they need to access them, can look in the git history at the corresponding tag?

Nicogene:
> Hi @Nicogene, I have only a doubt: I saw you updated all the templates file replacing embObjStrain w/ embObjFTSensor. So here is my question: do we want to replace also the old templates?
> 
> In my mind, the old templates are related to old versions, so we don't need to update them, just the current template.
> 
> What do you think?

Hi @valegagge, the idea is to remove `embObjStrain` (see https://github.com/robotology/icub-main/issues/875) since it works only w/ `analogServer` that will be removed soon, so those templates will not work anymore if we keep `embObjStrain`, but if it is for keeping an history I can undo the changes for the templates!

valegagge:
> Hi @valegagge, the idea is to remove embObjStrain (see https://github.com/robotology/icub-main/issues/875) since it works only w/ analogServer that will be removed soon, so those templates will not work anymore if we keep embObjStrain, but if it is for keeping an history I can undo the changes for the templates!

Hi @Nicogene ,
I think that the configuration files compliant with a template before the number 6, (that is current) don't work anymore because the parser and the module that interacts with such files are changed.
They can work with a particular set of revisions, but I think we lost track of this relation.

Anyway, I don't have a strong opinion on the matter of these changes, but I suggest reverting the template files before version 6.

What do you think?

pattacini:
>Anyway, I don't have a strong opinion on the matter of these changes, but I suggest reverting the template files before version 6.

Ok for this PR 👍🏻 

>I think that the configuration files compliant with a template before the number 6, (that is current) don't work anymore because the parser and the module that interacts with such files are changed.
They can work with a particular set of revisions, but I think we lost track of this relation.

Unless there are particular reasons, I would be in favor of removing the old templates outright[^1], keeping only the last one.

[^1]: They'll remain within Git history anyway.

valegagge:
> Unless there are particular reasons, I would be in favor of removing the old templates 

I do not completely agree: if a robot is running an old sw-ditro related to an old template, how does he/she verify the meaning of the parameter configuration if the templates are not available? 

What do you think if we move the old templates to a folder like `old-templates` in order to get the current template folder tidier?

pattacini:
>What do you think if we move the old templates to a folder like old-templates in order to get the current template folder tidier?

Fair enough 👍🏻

Nicogene:
>Anyway, I don't have a strong opinion on the matter of these changes, but I suggest reverting the template files before version 6.

I reverted the changes in the old templates, as soon as https://github.com/robotology/icub-main/pull/874 is merged, we can merge this PR.

pattacini:
CI passed ✅

pattacini (approved):

PR #524, opened 2023-05-22
ergoCub1.0 S/N000 – update left/right calib for startup and home position

sgiraz:
**What's new:**
- update  `startupPosition` and `homePosition` in `calibrators\left_arm-calib.xml` with the min values set for `jntPosMin` in `hardware\motorControl\left_arm-eb23-j7_10-mc.xml`
- update  `startupPosition` and `homePosition` in `calibrators\right_arm-calib.xml` with the min values set for `jntPosMin` in `hardware\motorControl\right_arm-eb22-j7_10-mc.xml`

**Note:**
- ⚠️ @AntonioConsilvio states that the `jntPosMin` parameter was set to 0 for the right arm since its joints can physically reach the 0 position. However, this is not the case for the left arm. Consequently, I adjusted the `jntPosMin` to 4 for the left arm to maintain consistency, considering it was already set to 4 for the right arm.
- Probably the `jntPosMin` and `jntPosMax` values for both hands may be improved to be more accurate 

cc @valegagge @MSECode @Nicogene

### DIFF
--- a/ergoCubSN000/calibrators/left_arm-calib.xml
+++ b/ergoCubSN000/calibrators/left_arm-calib.xml
@@ -12,7 +12,7 @@
 
 	<group name="HOME">
 	<!--                                shoulder-pitch    shoulder-roll    shoulder-yaw    elbow    wrist-yaw    wrist-roll    wrist-pitch    thumb_add    thumb_oc    index_add    index_oc    middle_oc    ring_pinky_oc -->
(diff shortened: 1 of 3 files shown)


pattacini (approved):

pattacini:
`jntPos[Min|Max]` refer to the SW bounds, whereas `hardwareJntPos[Min|Max]` are the HW bounds.
It's important the latter coincide with the HW, while the former can be more restrictive.

PR #525, opened 2023-05-22
Added 3DMID configuration files in experimentalSetups 

simeonedussoni:
In the 3DMID_fingertip_CAN fiolder there are all the files needed to launch the DAQ and visualization via yarpscope of the signals from the 3DMID fingertip. All is done through the CAN interface.


pattacini (approved):
Retargeted on `devel` as it's a new development.

pattacini:
Better off sticking to `3DMID_fingertip_CAN` rather than `3DMID_fingertip_Can`.
I'll amend it after merging.

PR #535, opened 2023-06-21
[ergocub] fixed FT FW version

pattacini:
This is a twin PR of #534.

cc @AntonioConsilvio


Nicogene (approved):

isorrentino:
Hi all with this PR ergoCubSN000 do not start since the firmware flashed on the strain of ergoCubSN000 is 2.1.0

Nicogene:
Hi @isorrentino, yes this change has been made to enforce the upgrade of the firmware to `2.2.0` since that version contains the fix on the IMU reference frame. We suggest to flash the latest strain2 firmware

PR #539, opened 2023-07-06
ergoCubSN000: tune up position PID of index and middle fingers

mfussi66:
This PR updates the PID controllers of the index and middle fingers of the ergoCub hands. The methodology involved linear identification and robust tuning after the addition of parametric uncertainties, bringing improvements both in error tracking and PWM noise. The Derivative component is set to zero, as it stiffens the system too much.


pattacini (approved):
See https://github.com/icub-tech-iit/ergocub-software/pull/143 for more details on the tuning methodology.

PR #553, opened 2023-08-10
fixed calibrator for CAN device

davidetome:
This PR fixes the calibrator for CAN devices on these robots : 

- iCubGenova06
- iCubNancy01

### DIFF
--- a/iCubGenova06/calibrators/head-calib.xml
+++ b/iCubGenova06/calibrators/head-calib.xml
@@ -2,7 +2,7 @@
 <!DOCTYPE devices PUBLIC "-//YARP//DTD yarprobotinterface 3.0//EN" "http://www.yarp.it/DTD/yarprobotinterfaceV3.0.dtd">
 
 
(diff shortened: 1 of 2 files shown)


pattacini (approved):
Thanks @davidetome 👍🏻

davidetome:
Hi @pattacini !

> I can still see the use of parametricCalibratorETH on the other parts of the robot iCubNancy01. Could you fix them as well?

The other parts are equipped w/ EMS so I thought that was right to leave Eth  , only the head is CAN based.
Isn't it?

>  Also, given that this is a fix, would you mind rebasing your PR against the upstream master instead of devel?
Sure

pattacini:
>The other parts are equipped w/ EMS so I thought that was right to leave Eth , only the head is CAN based.
Isn't it?

I don't know actually. I assumed it was a full CAN robot. If you confirm that the other parts are ETH based, that's ok!

pattacini (approved):
Just changing the base branch doesn't work as it brings other commits into play. Let me merge this against `devel` and then cherry-pick it into `master`.

PR #554, opened 2023-08-11
ergoCub000: Update right wrist calibration values

mfussi66:
This PR is necessary since the maintenance procedure required replacement of the right wrist motor.


mfussi66:
cc @pattacini @Nicogene @sgiraz

pattacini (approved):

PR #565, opened 2023-09-11
[iCubErzelli02] New configuration files

AntonioConsilvio:
## What changes:

This PR updates the configuration files after the iCubErzelli02 maintenance service.

cc @sgiraz


sgiraz:
Just to clarify, the whole robot can be launched using the `original_icub_all.xml` which is a sort of backup file. To simplify the startup process and the job for researchers in Erzelli, we agreed to remove the lower body from the startup (see `icub_all.xml`).

Nicogene (approved):

pattacini (approved):

PR #572, opened 2023-09-22
Add battery configuration files

MSECode:
This PR brings the following changes:
- Update the battery experimental setup
- Now the `experimentalSetups/battery` configuration can be used with:
    - only the BAT board connected to the EMS
    - only the BMS board connected to the EMS
    - both BAT and BMS boards connected to different EMS boards as for the `ergoCub` robot  

- Update test battery files to used both BMS and BAT streaming to different ports using the coupled battery bms+bat together


MSECode:
cc: @valegagge

pattacini (approved):

PR #574, opened 2023-10-04
Fix wrong coefficients coupling the eyes' vergence

### LINKED ISSUE #228: Wrong coefficients coupling the eyes' vergence?
Given that the vergence `Vg` amounts to the difference between the single eyes' joints `L-R` (see the [wiki][1]), I think that the [values contained in the eyes' coupling matrices][2] should be **`-1.0 1.0`** instead of `-0.5  0.5`.

Differently, as `Vs = (L + R) / 2` it turns out that the line above reporting `0.5  0.5` is spot on.

Can this reasoning be assumed correct?

This is somehow linked to https://github.com/robotology/icub-firmware/issues/71.

cc @davidetome @randaz81 @ale-git @traversaro @plinioMoreno 

[1]: https://wiki.icub.org/wiki/Vergence,_Version_and_Disparity
[2]: https://github.com/robotology/robots-configuration/blob/master/iCubGenova04/hardware/mechanicals/head-eb21-j2_5-mec.xml#L45

martinaxgloria:
This PR fixes #228. 

I did some tests on `iCubGenova11` with this change and I posted [here](https://github.com/robotology/robots-configuration/issues/228#issuecomment-1746370597) the results. Then, I proceeded with the update of the other maintained robots.


pattacini (approved):
Thanks heaps @martinaxgloria!

What I found a bit puzzling is how we handle the coupling for those old CAN robots that we still maintain, such as `iCubNancy01`. It doesn't seem to have this info in its folders.

cc @Nicogene @davidetome

martinaxgloria:
Yes, @pattacini. I noticed that those robots don't have this kind of information in the configuration files, so I skipped them because I didn't find the references anywhere. Maybe @davidetome could have the answer

pattacini:
Checked w/ @davidetome @Nicogene and finally @maggia80 👍🏻 
It turned out that with old robots the coupling is hand-coded in the FW as per https://icub-tech-iit.github.io/documentation/icub_firmware/associated-firmware/associated-firmware/.

martinaxgloria:
Great, thank you all!

PR #586, opened 2023-11-08
ergoCubSN001: fix cluster-config.xml

Nicogene:
As per title

cc @GiulioRomualdi @lrapetti


pattacini (approved):
Thanks for the fix.
I'll be cherry-picking it on `master`.

PR #598, opened 2023-12-05
[ergoCubSN001] Enable lowerarm and more

AntonioConsilvio:
## What changes:

This PR enables lowerarm (although wrists have some problems, it was chosen to keep them enabled). 
It also disables AMO in some joints due to malfunctions (will try to fix them as soon as possible).

In addition, there are also useful changes to the AMI demos (@GiulioRomualdi).

## Note 

The robot has been tested and works properly with this configuration.

### DIFF
--- a/ergoCubSN001/calibrators/left_arm-calib.xml
+++ b/ergoCubSN001/calibrators/left_arm-calib.xml
@@ -12,7 +12,7 @@
 
 	<group name="HOME">
 	<!--                                shoulder-pitch    shoulder-roll    shoulder-yaw    elbow    wrist-yaw    wrist-roll    wrist-pitch    thumb_add    thumb_oc    index_add    index_oc    middle_oc    ring_pinky_oc -->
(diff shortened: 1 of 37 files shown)


GiulioRomualdi:
The only issue with this PR is that we had to remove the arms FTs from WBD since @AntonioConsilvio informed us that the sensors were continuously rebooting. In order to enable whole-body dynamics, we manually removed it from the URDF model.

In conclusion, this PR introduces a WBD configuration that is not compatible with the SN001 model pushed in ergocub-software

pattacini:
Thanks @AntonioConsilvio for the PR and thanks @GiulioRomualdi for your note on WBD.

I would proceed with emerging the PR anyway and asking @Nicogene to take care of updating the URDF model in `ergocub-software` if you all agree.

I'm open to alternatives though.

GiulioRomualdi:
Fine with me :)

pattacini (approved):

PR #603, opened 2023-12-27
Feature/fw updater bat

MSECode:
This PR brings the following changes:
setting the configuration of the BAT CAN Address to CANId ==> 2 since we are using specific messages that are of class 6 and the sending address is specifically 2, such as 620 and 628.

So user must always set the CANId for the BAT to 2.


valegagge (approved):
LGTM

valegagge (approved):
LGTM

valegagge (approved):
LGTM

pattacini (approved):

PR #609, opened 2024-01-15
[iCubGenova11] Publish Euler angles from IMU on rfe

martinaxgloria:
This PR adds the possibility to stream the orientation data from the IMU mounted on the rfe.

cc @Nicogene


pattacini:
Awaiting @Nicogene's review, who worked right on this.

Nicogene (approved):

PR #615, opened 2024-02-15
[iCubGenova11] Add config files for FT IMUs of the arms

martinaxgloria:
This PR aims to add the configuration files for reading from the IMU of the FT sensors mounted on the left and right arms. Moreover, I added also an `alljoints-inertials_wrapper.xml` and `alljoints-inertials_remapper.xml` to wrap up the inertials with a `multipleanalogsensorsremapper`. 

I didn't add those files in the `icub_all_no_legs.xml` which is the startup xml of the robot since I don't know if someone is interested in that information. 

cc @Nicogene @traversaro @pattacini


pattacini (approved):
Thanks!

PR #616, opened 2024-02-16
[ergoCubSN001] Add a head-mc network wrapper a lower streaming frequency

GiulioRomualdi:
cc @S-Dafarra


pattacini (approved):
Hi @GiulioRomualdi @S-Dafarra 

Do you think that updating `ergoCubSN000` as well could be worth it?

pattacini:
Merging anyway.
`ergoCubSN000` can be updated afterward, in case.

PR #634, opened 2024-03-26
[ergoCubSN001] Change the version of the BAT and BMS

GiulioRomualdi:
I found these files uncommitted on the robot. I think they are important.

cc @MSECode @traversaro @pattacini

### DIFF
--- a/ergoCubSN001/hardware/battery/battery_bat.xml
+++ b/ergoCubSN001/hardware/battery/battery_bat.xml
@@ -20,8 +20,8 @@
                     <param name="minor">            0                   </param>
                 </group>
                 <group name="FIRMWARE">
-                    <param name="major">            1                   </param>
-                    <param name="minor">            3                   </param>
+                    <param name="major">            0                   </param>
+                    <param name="minor">            0                   </param>
                     <param name="build">            0                   </param>
                 </group>
             </group>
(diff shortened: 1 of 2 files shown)


traversaro:
cc @valegagge

MSECode:
Hi @GiulioRomualdi,
I think the version should be aligned with what defined in the firmware info file at these lines: https://github.com/robotology/icub-firmware-build/blob/fb6d2150403efcae8ffede9e8887a5473997e2df/info/firmware.info.xml#L48-L49 
It is possible to set all zeros to the versions in order to skip the checks about the matching between what defined in the conf and what running on the board at the YRI startup but I'm not sure if it is the best practice to use. I think we should set the version as the one running on the board, i.e. 1.3.3. 
Anyways, is the YRI rising any error flags complaining about the version mismatch?

pattacini:
Hi @MSECode 

>It is possible to set all zeros to the versions in order to skip the checks about the matching between what defined in the conf and what running on the board at the YRI startup but I'm not sure if it is the best practice to use. I think we should set the version as the one running on the board, i.e. 1.3.3.

We recently decided to drop this versioning done inside the XML files because it is barely maintainable.
New developments on the FW side will compensate for this.

pattacini (approved):
Thanks @GiulioRomualdi

MSECode:
> Hi @MSECode
> 
> > It is possible to set all zeros to the versions in order to skip the checks about the matching between what defined in the conf and what running on the board at the YRI startup but I'm not sure if it is the best practice to use. I think we should set the version as the one running on the board, i.e. 1.3.3.
> 
> We recently decided to drop this versioning done inside the XML files because it is barely maintainable. New developments on the FW side will compensate for this.

Oks, good to know. I did not know about that. So I think it's all good with that PR and thanks @GiulioRomualdi

Issue #636, opened 2024-03-28
IMU sensors: alignment real ergoCub/iCub w/ simulated one


Nicogene:
This PR:
- https://github.com/robotology/robots-configuration/pull/632

Added the remapper for the imus that uses `sensorName`s that are not coherent with the actual sensor names of the exposed imu except for `iCubGenova11`.

`embObjIMU` exposes as sensor name the `id` if `sensorName` is not specified, and `sensorName` is present in only the `iCubGenova11` imus xml.

Moreover, I noticed that on urdf side, iCub and ergoCub use different names for the imus:

- https://github.com/icub-tech-iit/ergocub-software/blob/4a7ce3a1472b16c7a3372623199c3ccbaa761a3d/urdf/creo2urdf/data/ergocub1_1/ERGOCUB_all_options_gazebo.yaml#L1166
- https://github.com/robotology/icub-models-generator/blob/5c6e49cfd9e96c9d118ec2c16f98b9e45e4b379d/simmechanics/data/icub2_5/ICUB_2-5_BB_simmechanics_options.yaml.in#L693

## Mandatory
- @traversaro Stakeholder need: The sensor name used on simulator and real robot (either for iCub and ergoCub) must be coherent
- `sensorName` must be added in all FT imu xml.

## Nice to have
- iCub and ergoCub uses the same sensor names

## Nicer to have
- Every sensor follows this guideline (or we can change it as we please)
	- https://icub-tech-iit.github.io/documentation/robot_sensors/ergocub_sensors/ 

This activity can be done in more steps, the mandatory points are the one more urgent, but they depend on if we want to "cleanup" the sensor name situation 

cc @pattacini

traversaro:
fyi @robotology/iit-ami-scrum-masters when implemented this may affect configuration files on our side.

martinaxgloria:
As discussed in a private chat with @Nicogene, we decided to add the `sensorName` tag in all the FT IMUs XML files and the chosen names must be coherent with the simulated ones.

- For the ergoCub side, since:
 	- https://github.com/icub-tech-iit/ergocub-software/pull/217
	- https://github.com/icub-tech-iit/ergocub-software/pull/232

	the sensorName tag should be `<part>_ft_imu` (for example, `l_arm_ft_imu`) to make it different from the FT ones without the suffix `_imu`:

[...]

PR #640, opened 2024-04-11
IMU sensors - Include alljoints-inertials_wrapper/remapper in YRI xml files

martinaxgloria:
As pointed out by @traversaro [here](https://github.com/icub-tech-iit/ergocub-software/pull/232#issuecomment-2039771072), the  files that expose via multipleanalogsensorsremapper the orientation measurements of the available IMUs are now included in the XML used by the yarprobotinterface

This PR will complete https://github.com/robotology/robots-configuration/issues/636

### DIFF
--- a/ergoCubSN000/ergocub_all.xml
+++ b/ergoCubSN000/ergocub_all.xml
@@ -106,7 +106,9 @@
     <!-- FT SENSORS - IMU -->
     <xi:include href="hardware/inertials/left_arm-eb2-j0_1-inertial.xml" />
     <xi:include href="hardware/inertials/right_arm-eb1-j0_1-inertial.xml" />
+    <xi:include href="hardware/inertials/left_leg-eb8-j0_3-inertial.xml" />
     <xi:include href="hardware/inertials/left_leg-eb9-j4_5-inertial.xml" />
+    <xi:include href="hardware/inertials/right_leg-eb6-j0_3-inertial.xml" />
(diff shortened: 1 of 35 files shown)


pattacini (changes requested):
Hi @martinaxgloria 
CI is failing.

martinaxgloria:
Probably I'm missing some configuration files, I'll take a look into

martinaxgloria:
As expected, the CI was falling due to some missing configuration files. Moreover, I noticed that the dry-run checks only the XML that the YRI uses by default, that is the one under `yarprobotinterface.ini`. Since for the other XML there's no CI check (for example, for `ergocub_wbd.xml` or for `ergocub_all_ros2.xml`, but also some iCubs have more than one YRI launch file), what about adding the alljoints-inertials_wrapper/remapper only in the main yarprobotinterface XML file? Or are there other files that are used and that could be worth manually maintaining them? 

cc @traversaro @Nicogene @pattacini

traversaro:
Ok, I think I did not get your question.

martinaxgloria:
Aligned with @traversaro t2t, we decided to expose IMU measurements only in the yri default xml for the iCubs, while for ergoCubSN001 and ergoCubSN002 in all `ergocub_*.xml`.

cc @Nicogene @pattacini

pattacini (approved):
Thanks heaps for the fix, @martinaxgloria 🚀

PR #642, opened 2024-04-17
[ergoCubSN002] Updating configuration files after robot testing

AntonioConsilvio:
## What changes:

This PR updates the ergoCubSN002 configurations files for the official robot release.

## Note:

- The temperature sensors (TP100) of three joints are disabled due to malfunctioning: Torso Roll, Torso Yaw and left shoulder roll.

- The sign of the Torso Pitch PIDs is back as it was originally. So, the opposite of ergoCubSN001, where they were reversed during the initial testing phase.

cc @S-Dafarra @GiulioRomualdi

### DIFF
--- a/ergoCubSN002/calibrators/head-calib.xml
+++ b/ergoCubSN002/calibrators/head-calib.xml
@@ -20,7 +20,7 @@
         <!-- joint name                         neck-pitch  neck-roll   neck-yaw    eyes-tilt -->
         <group name="CALIBRATION">
             <param name="calibrationType">      12          12          12          12         </param>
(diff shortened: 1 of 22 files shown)


pattacini (approved):
Thanks!

>The sign of the Torso Pitch PIDs is back as it was originally. So, the opposite of ergoCubSN001, where they were reversed during the initial testing phase.

It was due to a miscabling?

AntonioConsilvio:
@pattacini probability, the rotor had the phases mounted in reverse internally. To prevent these unexpected things from happening again, we included in the procedure for testing the rotors (before mounting them on the robot) the test to check that they move clockwise or anticlockwise given a positive voltage.

PR #645, opened 2024-04-21
Cleanup yarprobotinterface conf files for ergoCub

traversaro:
Fix https://github.com/robotology/robots-configuration/issues/641 . 

The changes in the PR are the one described in https://github.com/robotology/robots-configuration/issues/641#issue-2239652775 . I actually used the same ergocub.xml for all ergoCub robots, with the following differences:
* Each robot, has a different `robot name=` attribute.
* In `ergoCubSN00`, the following lines are commented:
~~~
    <!-- BATTERY BAT -->
    <!-- <xi:include href="wrappers/battery/battery_bat.xml" /> -->
    <!-- <xi:include href="hardware/battery/battery_bat.xml" /> -->
~~~
I am not sure why this happening, but to avoid regression I left these lines commented also in the new `ergocub.xml`
* In `ergoCubSN001`, there was also a `<xi:include href="wrappers/motorControl/head-mc-slow_wrapper.xml" />` line that is not in other robots. Is this actually used? For the time being, I did not included that in the PR.

To avoid regressions, this PR does not remove any existing `ergocub_***.xml` files. Once the new `ergocub.xml` are tested and working, we can remove the `ergocub_***.xml` files in a follow up PR.

PR #648, opened 2024-04-22
Disable parts of ergoCub relying on `disable_by` option

pattacini:
Follow-up of #645 #647


traversaro:
I guess both ROS2 and wholebodynamics should be disabled if any part of the robot is disabled, as they currently hardcode the list of joints they consider.

pattacini:
Good point @traversaro 👍🏻 
I've updated the files accordingly.

traversaro (approved):

PR #652, opened 2024-05-16
ergoCubSN001 - update network file

davidetome:
This PR updates the network file of the robots ergoCub SN001 and ergoCubSN002

cc @valegagge

### DIFF
--- a/ergoCubSN001/network.ergoCubSN001.xml
+++ b/ergoCubSN001/network.ergoCubSN001.xml
@@ -53,17 +53,117 @@
         <board type='mc4plus' name="head-eb21-j2_3">
             <ondevice>ETH</ondevice>
             <ataddress ip="10.0.1.21" />
-            <connected bus="ETH" prev="10.0.1.20" next="10.0.1.22" />
+            <connected bus="ETH" prev="10.0.1.20" next="10.0.1.2" />
         </board>
 
     </part>
 
     <part name="left_arm">
 
-        <board type='ems4' name="left_arm-eb1-j0_1">
+        <board type='ems4' name="left_arm-eb2-j0_1">
+            <ondevice>ETH</ondevice>
+            <ataddress ip="10.0.1.2" />
+			<connected bus="ETH" prev="10.0.1.21" next="10.0.1.4" />
+        </board>
+
+       <board type='foc' name="foc.1.1">
+            <ondevice>ETH</ondevice>
+            <ataddress ip="10.0.1.2" canbus="1" canadr="1"  />
+            <connected bus="CAN" />
+        </board>
+
+        <board type='foc' name="foc.1.2">
+            <ondevice>ETH</ondevice>
+            <ataddress ip="10.0.1.2" canbus="1" canadr="2"  />
+            <connected bus="CAN" />
+        </board>
+		
+        <board type='strain2c' name="strain2.13">
+            <ondevice>ETH</ondevice>
+            <ataddress ip="10.0.1.2" canbus="2" canadr="13"  />
+            <connected bus="CAN" />
+        </board>
+
+         <board type='ems4' name="left_arm-eb4-j2_3">
+            <ondevice>ETH</ondevice>
+            <ataddress ip="10.0.1.4" />
+			<connected bus="ETH" prev="10.0.1.2" next="10.0.1.23" />
+        </board>       
+
+        <board type='foc' name="foc.2.1">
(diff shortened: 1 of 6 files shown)


ergoCubSN001/network.ergoCubSN001.xml
```diff
@@ -227,130 +286,141 @@
...
+        </board>
+
         <board type='foc' name="foc.7.1">
             <ondevice>ETH</ondevice>
             <ataddress ip="10.0.1.7" canbus="1" canadr="1"  />
             <connected bus="CAN" />
         </board>
 
-        <board type='foc' name="foc.7.2">
+        <board type='foc' name="foc.7`.2">
```
pattacini, line 339: There could be a typo here.

pattacini (changes requested):
Hi @davidetome 

Could you provide some insights underlying this big change?
Also, would it be worth covering the other ergoCub as well?

Possibly, there's one typo requiring attention. Didn't spot others but, please, check it.

davidetome:
> Hi @davidetome
> 
> Could you provide some insights underlying this big change? Also, would it be worth covering the other ergoCub as well?
> 
> Possibly, there's one typo requiring attention. Didn't spot others but, please, check it.

Ciao @pattacini !!!
 
   I noticed that the [network ](https://github.com/robotology/robots-configuration/blob/devel/ergoCubSN001/network.ergoCubSN001.xml)file was not aligned to the real [architecture ](https://github.com/icub-tech-iit/electronics-wiring/blob/master/ergocub1/ergocub1.2/pdf/ergoCub1.2_Logic_17929_1.2.0.pdf),

I just fixed some errors and checked again w/ the schematic. 

Also SN002 is not correct (probably copied when the folder was created)

cc @valegagge @AntonioConsilvio

pattacini:
> I noticed that the [network ](https://github.com/robotology/robots-configuration/blob/devel/ergoCubSN001/network.ergoCubSN001.xml)file was not aligned to the real [architecture ](https://github.com/icub-tech-iit/electronics-wiring/blob/master/ergocub1/ergocub1.2/pdf/ergoCub1.2_Logic_17929_1.2.0.pdf),
> 
> I just fixed some errors and checked again w/ the schematic.

Superb! Thanks heaps!


> Also SN002 is not correct (probably copied when the folder was created)

Then, it'd be great if we could fix it up as well.

@AntonioConsilvio, you may consider checking these things out for SN000, which is by us these days.

davidetome:
@pattacini @AntonioConsilvio , as suggested by @valegagge , I also fixed SN002 with the last commit 👍🏼

pattacini:
Thanks @davidetome 🚀

pattacini (approved):
Awaiting the PR to be put in ready for review.

AntonioConsilvio:
Hi! Well done @davidetome.

I'd just like to add that some boards are actually not connected to each other.

For example, the `10.0.1.2` board, has no boards before it, because it's connected to the ETH switch (Gigablox) so it might make sense to edit line 66 in exactly the same way as 10.0.1.5 (i.e. without the `prev` parameter - line 266)? 
Similarly, 10.0.1.21 has no boards after it, so it might make sense to remove the `next` parameter?

Other boards might have this little problem, in the architecture the ETH switch is represented like this:

image

cc @pattacini

davidetome:
> Hi! Well done @davidetome.
> 
> I'd just like to add that some boards are actually not connected to each other.
> 
> For example, the `10.0.1.2` board, has no boards before it, because it's connected to the ETH switch (Gigablox) so it might make sense to edit line 66 in exactly the same way as 10.0.1.5 (i.e. without the `prev` parameter - line 266)? Similarly, 10.0.1.21 has no boards after it, so it might make sense to remove the `next` parameter (line 56)?
> 
> Other boards might have this little problem, in the [architecture](https://github.com/icub-tech-iit/electronics-wiring-public/blob/master/ergocub1/ergocub1.2/pdf/ergoCub1.2_Logic_17929_1.2.0.pdf) the ETH switch is represented like this:
> 
> image
> 
> cc @pattacini

ciao @AntonioConsilvio , it doesn't matter...the previous and next tag are only for info when reading the file.
you can also not declare them and the script runs  the same

davidetome:
@AntonioConsilvio , just to make you happy I just fixed them with the last commit 😄  👍🏼

AntonioConsilvio (approved):

valegagge (approved):
LGTM

PR #667, opened 2024-07-08
Fixed typo in local-workflow-tech

S-Dafarra:
The remote to change is not ``origin`` but ``icub-tech``

### DIFF
--- a/.github/local-workflow-tech.md
+++ b/.github/local-workflow-tech.md
@@ -34,7 +34,7 @@ Technicalities underlying the fast-pace local workflow
 > [!tip]
 > Instead of relying on the credential helper, one can resort to a simpler method:
 > ```console
-> git remote set-url origin https://x-access-token:${PAT}@github.com/icub-tech-iit/robots-configuration.git
+> git remote set-url icub-tech https://x-access-token:${PAT}@github.com/icub-tech-iit/robots-configuration.git
 > ```
 
 ## 🔲 Ensure that the author information is specified explicitly at commit time


pattacini:
Actually, the convention for remotes on a fork usually is:
- `origin`: points to the fork repo on GH.
- `upstream`: points to the upstream repo on GH.

I wasn't aware there was a remote called `icub-tech`, which would be peculiar to our system and not created at clone time.

S-Dafarra:
> Actually, the convention for remotes on a fork usually is:
> 
> * `origin`: points to the fork repo on GH.
> * `upstream`: points to the upstream repo on GH.
> 
> I wasn't aware there was a remote called `icub-tech`, which would be peculiar to our system and not created at clone time.

Actually, the repo gets cloned by the super build using robotology as origin. I thought it would be less error prone specifying a different remote name for the robot workflows only. This also enforces the good practice of not changing origin when adding a new fork to the repo

pattacini:
>Actually, the repo gets cloned by the super build using robotology as origin.

I see, I missed this point.
Then, I would explain somewhere in the instructions that `icub-tech` is the remote pointing to the fork.

pattacini:
> Then, I would explain somewhere in the instructions that `icub-tech` is the remote pointing to the fork.

It's already introduced here:
- https://github.com/robotology/robots-configuration/blob/2847adc1f34afc4c79c640cac567871073479b7b/.github/local-workflow-tech.md?plain=1#L15

pattacini (approved):

PR #670, opened 2024-07-18
Devel ergo cub sn002

pattacini:
Integrating upstream the ergoCubSN002 changes.

cc @S-Dafarra


pattacini:
To integrate upstream the changes, we need to solve those conflicts.
As a result, the fork branch will get modified and a `git pull --rebase` will need to be issued on the ergoCubSN002 system.

cc @S-Dafarra

PR #675, opened 2024-08-07
Notify YRI about the availability of disable/enable tags

pattacini:
cc @traversaro


traversaro:
From what I understand, this change will enable ros2 by default, and disable by default all the parts of the robot and wholebodydynamics, so I guess I am missing something?

PR #677, opened 2024-09-06
Set `estimateJointVelocityAcceleration` for ergoCubSN00*

### LINKED ISSUE #668: wholeBodyDynamics configuration files for ergoCubSN00* not setting the estimateJointVelocityAcceleration option?
While working on https://github.com/robotology/whole-body-estimators/pull/194 , I noticed that the test was printing a warning:

~~~
[WARNING] wholeBodyDynamics: estimateVelocityAccelerationOptionName bool parameter missing, please specify it.";
[WARNING] wholeBodyDynamics: setting estimateVelocityAccelerationOptionName to the default value of true, but this is a deprecated behaviour that will be removed in the future.";
~~~

that was surprising for me, as I based the test introduced in https://github.com/robotology/whole-body-estimators/pull/177 on the configurations of ergoCubSN00* . Indeed, by looking in the configuration files for ergoCubSN00*, I noticed that the `estimateJointVelocityAcceleration` option is never set (the error refers to a different name, but it is a typo).

**This means that for ergoCubSN00* torque estimation we never used the velocity estimated at low level, but always the velocity estimated at high level with a kalman filter.**

I guess we should get rid of this warning by explicitly setting `estimateJointVelocityAcceleration` (as done for iCub3 in https://github.com/robotology/robots-configuration/blob/8883a01d17fbb6da01372843ede67b066ce91ffd/robots-icebox/iCubGenova09/estimators/wholebodydynamics.xml#L23) and then we can modify wbd to raise an error if the parameter is not set.

fyi @GiulioRomualdi @isorrentino @loremoretti

pattacini:
Fix #668.

### DIFF
--- a/ergoCubSN000/estimators/wholebodydynamics.xml
+++ b/ergoCubSN000/estimators/wholebodydynamics.xml
@@ -20,6 +20,7 @@
         <param name="devicePeriodInSeconds">0.002</param>
         <param name="useSkinForContacts">false</param>
         <param name="publishNetExternalWrenches">true</param>
+        <param name="estimateJointVelocityAcceleration">false</param>  <!-- if true a kf will estimate the joint velocities and accelerations -->
 	<!--<param name="assume_fixed">root_link</param>-->
 	<group name="HW_USE_MAS_IMU">
 	    <param name="accelerometer">waist_imu_0</param>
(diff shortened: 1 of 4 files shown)


traversaro (approved):

traversaro:
Thanks @pattacini !

PR #684, opened 2024-10-10
ergoCubSN001-002: Update amcbdlc and amc versions

mfussi66:
In relation to the latest PRs in icub-firmware (https://github.com/robotology/icub-firmware/pull/523) and icub-firmware-build (https://github.com/robotology/icub-firmware-build/pull/165), this PR updates the boards as per title:

AMC2C : 3.1
AMCBLDC: 2.1

### DIFF
--- a/ergoCubSN001/hardware/motorControl/left_arm-eb31-j4_6-mc_service.xml
+++ b/ergoCubSN001/hardware/motorControl/left_arm-eb31-j4_6-mc_service.xml
@@ -21,8 +21,8 @@
                 </group>
                 <group name="FIRMWARE">
                     <param name="major">            2           3       </param>
-                    <param name="minor">            0           0       </param>
-                    <param name="build">            13          3       </param>
+                    <param name="minor">            1           1       </param>
+                    <param name="build">            0           0       </param>
                 </group>
             </group>
 
--- a/ergoCubSN001/hardware/motorControl/right_arm-eb30-j4_6-mc_service.xml
+++ b/ergoCubSN001/hardware/motorControl/right_arm-eb30-j4_6-mc_service.xml
@@ -21,8 +21,8 @@
                 </group>
                 <group name="FIRMWARE">
                     <param name="major">            2           3       </param>
-                    <param name="minor">            0           0       </param>
-                    <param name="build">            13          3       </param>
+                    <param name="minor">            1           1       </param>
+                    <param name="build">            0           0      </param>
                 </group>
             </group>
 
--- a/ergoCubSN002/hardware/motorControl/left_arm-eb31-j4_6-mc_service.xml
+++ b/ergoCubSN002/hardware/motorControl/left_arm-eb31-j4_6-mc_service.xml
@@ -21,8 +21,8 @@
                 </group>
                 <group name="FIRMWARE">
                     <param name="major">            2           3       </param>
-                    <param name="minor">            0           0       </param>
-                    <param name="build">            13          3       </param>
+                    <param name="minor">            1           1       </param>
+                    <param name="build">            0           0       </param>
                 </group>
             </group>
 
--- a/ergoCubSN002/hardware/motorControl/right_arm-eb30-j4_6-mc_service.xml
+++ b/ergoCubSN002/hardware/motorControl/right_arm-eb30-j4_6-mc_service.xml
@@ -21,8 +21,8 @@
                 </group>
                 <group name="FIRMWARE">
                     <param name="major">            2           3       </param>
-                    <param name="minor">            0           0       </param>
-                    <param name="build">            13          3       </param>
+                    <param name="minor">            0           1       </param>
+                    <param name="build">            13          0       </param>
                 </group>
             </group>
 
--- a/ergoCubSN002/hardware/motorControl/right_arm-eb30-j4_6-mc_service.xml
+++ b/ergoCubSN002/hardware/motorControl/right_arm-eb30-j4_6-mc_service.xml
@@ -21,8 +21,8 @@
                 </group>
                 <group name="FIRMWARE">
                     <param name="major">            2           3       </param>
-                    <param name="minor">            0           1       </param>
-                    <param name="build">            13          0       </param>
+                    <param name="minor">            1           1       </param>
+                    <param name="build">            0          0       </param>
                 </group>
             </group>
 


pattacini:
Hi @mfussi66 

I've seen that the PR is still draft, so I imagine you'll be doing some tests.
Hence, what's your strategy to test these changes on the robot?

Ideally, given the small updates, you would be testing the changes on top of the robot-specific fork branches:
- https://github.com/icub-tech-iit/robots-configuration/tree/devel-ergoCubSN001
- https://github.com/icub-tech-iit/robots-configuration/tree/devel-ergoCubSN002

This guarantees that everything remains consistent on the robot setups.

Those changes will be then propagated back to upstream (aka here) with periodical PRs. Or, we can do that manually straight away.

I'm bringing up this point to get everyone aligned, especially considering that these are the first times we've tested this approach.

cc @S-Dafarra @traversaro

mfussi66:
> Hence, what's your strategy to test these changes on the robot?

The tests were performed only on ergoCubSN000 (https://github.com/robotology/icub-firmware-models/issues/92) and ergoCubSN001 (https://github.com/robotology/icub-firmware-models/issues/95)

On SN000 since the AMC does not perform motor control, I only tested the AMCBLDCs. on SN001 I tested both the boards. 

SN002 is unavailable at the time of this PR, so i was not able to perform tests, but the boards' features should be the same as SN001, hence why this PR targets that one, too.

However, it does not target SN000 since on that robot no check is performed on the AMCBLDC firmware's version, and I did not want to disrupt its operation mode by enforcing a version.

The configuration files I used are stored in my fork and the yarprobotinterface boots up only the wrist boards.

Let me know if this workflow can be considered equivalent to the contributing guidelines, and if this PR can be merged. 

I'll mark this as ready for review since the tests were performed, and put it back to draft if deemed insufficient.

pattacini:
The most important point is:
- How did you leave the robot?
  I imagine you flashed back to the previous version of the FW once you finished testing. Is that correct?

If so, I think we can merge the PR and once AMI decides to align the fork with upstream the boards will need to be updated. This operation is usually performed outright at each Distro release unless there are particular requests.

>The configuration files I used are stored in my fork and the yarprobotinterface boots up only the wrist boards.

The tests on the robot should be carried out on the complete setup to avoid underestimating any possible side effects, meaning launching the whole robot with the complete set of conf files.

---

Regarding this

>However, it does not target SN000 since on that robot no check is performed on the AMCBLDC firmware's version, and I did not want to disrupt its operation mode by enforcing a version.

I have a memory of the past when we decided to disable the FW versions, cc @GiulioRomualdi.
Apparently, this is done for SN000 but not for SN001/2.

mfussi66:
>    How did you leave the robot?
>   I imagine you flashed back to the previous version of the FW once you finished testing. Is that correct?

Yes, the firmware was restored to the version present in icub-firmware-build contained in the robotology-superbuild folder. 

So a rebase to devel would require a firmware flash.

> The tests on the robot should be carried out on the complete setup to avoid underestimating any possible side effects, meaning launching the whole robot with the complete set of conf files.

Agreed, at the time of testing the robots were not fully functional however, and I would have had to wait for repairs.

pattacini:
>Yes, the firmware was restored to the version present in icub-firmware-build contained in the robotology-superbuild folder.

Good 👍🏻 
So, let's wait a bit for others to chime in before merging.

>Agreed, at the time of testing the robots were not fully functional however, and I would have had to wait for repairs.

Clear 👍🏻

pattacini (approved):

pattacini:
Sure!
Awaiting that https://github.com/robotology/icub-firmware-build/pull/165 is merged before proceeding here.

marcoaccame (changes requested):
About tests: I am confident to say that since @mfussi66 has tested the binaries of the `amc2c` and `amcbldc` on two robots, that is enough to assume that they will work also on the third robot.

So, I am expecting to see six files changed: the `-mc_service.xml` for left and right wrists for the three `ergoCub` robots.


@mfussi66 I can see only three files changed, so you have probably missed the others. 

About merging: this PR has two other associated PRs that are ready to be merged.
- https://github.com/robotology/icub-firmware/pull/523
- https://github.com/robotology/icub-firmware-build/pull/165

We shall surely merge them together to this one, so that we don't have binaries un-aligned with the xml files.

pattacini (changes requested):
@marcoaccame is right.
The right arm of SN002 is missing. @mfussi66 can you update it?

SN000 has version checking disabled as we reported in https://github.com/robotology/robots-configuration/pull/684#issuecomment-2404715109

mfussi66:
Thanks for noticing @marcoaccame and @pattacini , the last commits update the missing file.

marcoaccame (approved):

pattacini (approved):

pattacini:
Thanks @mfussi66 👍🏻 

>Would it be possible to wait for a release of icub-firmware-build before merging? In this way, when rebasing on top of devel, we also know that we have to update to a new release.

@S-Dafarra, the corresponding PR on `icub-firmware-build` has been merged, so we're good to go here as well.

We generally don't produce release of the build until the new Distro. I think you're ok with rebasing on top of devel.

S-Dafarra:
> We generally don't produce releases of the build until the new Distro. I think you're ok with rebasing on top of devel

Is it a hard dependency? I would like to remain consistent with all the FW of all the boards for a given version of icub-firmware-build. In other words, if we have to update some boards, I would rather update all the boards, otherwise it gets messy. The update usually goes smooth for a release, while for commits it might generate a domino effect with the yarp and icub-main versions. If this cannot be satisfied, I am good in merging this, but I would postpone the rebasing of the robots configuration branches until the next release. Does this sound reasonable?

pattacini:
I think that, in this case, it's safe to update only the boards controlling the wrist of the ergoCub robot. I'm referring here to 2 AMC and 4 AMCBLDC boards handling the left and right wrist.

Tagging `icub-firmware-build` means that tags should propagate upward, up to a given extent.

> I am good in merging this, but I would postpone the rebasing of the robots configuration branches until the next release. Does this sound reasonable?

I would like to merge this PR to remain consistent with the `devel` of our whole codebase.
When you aim to update the fork branches is not really critical for us, as we're happy with our tests, I'd say.

While waiting for the next Distro, if important devs/patches applied to devel will need to land on the physical robots, we will help with the tests/update.

Calling for feedback on this proposal, @S-Dafarra @mfussi66 @marcoaccame @Nicogene

S-Dafarra:
> I think that, in this case, it's safe to update only the boards controlling the wrist of the ergoCub robot. I'm referring here to 2 AMC and 4 AMCBLDC boards handling the left and right wrist.

I would discourage this approach. We have done it a few times in the past and it was never a good situation, ending up in unexpected results. One common assumption is that the robot is aligned to the fw version of icub-firmware-build status in the head/torso. If this is not the case, we might end up in unreproducible situations.

PR #690, opened 2024-11-28
Devel ergocub sn000

pattacini:
Alignment in view of distro 2024.11.0.


pattacini:
I think this is mergeable flawlessly.
cc @S-Dafarra @GiulioRomualdi

PR #692, opened 2024-11-28
Devel ergocub sn002

pattacini:
Alignment in view of distro 2024.11.0.

PR #693, opened 2024-12-03
[ergoCubSN002] Add missing amc2c boards in network.ergoCubSN002.xml

traversaro:
See https://github.com/robotology/icub-firmware-build/issues/176#issuecomment-2514964433 for details. I am not sure this is the correct solution, but I guess opening a PR with a concrete proposal for a change would make it easier to discuss about this.

@valegagge @marcoaccame @AntonioConsilvio @MSECode @S-Dafarra

### DIFF
--- a/ergoCubSN002/network.ergoCubSN002.xml
+++ b/ergoCubSN002/network.ergoCubSN002.xml
@@ -102,7 +102,7 @@
             <connected bus="CAN" />
         </board>
 
-	    <board type='mtb4' name="mtb4.1">
+	<board type='mtb4c' name="mtb4.1">
             <ondevice>ETH</ondevice>
             <ataddress ip="10.0.1.4" canbus="2" canadr="1"  />
             <connected bus="CAN" />
@@ -138,7 +138,7 @@
 			<connected bus="ETH" prev="10.0.1.25"/>
         </board>
 
-        <board type='amc2c' name="left_arm-eb31-j4_6">
+        <board type='amc' name="left_arm-eb31-j4_6">
             <ondevice>ETH</ondevice>
             <ataddress ip="10.0.1.31" />
 			<connected bus="ETH" prev="10.0.1.25"/>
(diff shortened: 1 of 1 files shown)


pattacini:
Hi @traversaro 

It seems that this PR is somehow already subsumed by #692 for what regards the `mtb4c`.
Nonetheless, #692 says nothing about `amc` vs. `am2c`.

S-Dafarra:
Discussing with @AntonioConsilvio, it turned out we had to use **both** fw, since the board is dual core. We had to flash both of them one after the other

pattacini:
Hi @S-Dafarra 

What does this mean in terms of XML files? That we have to give it a run first with `amc` in the XML and then replace it with `amc2c`?

valegagge:
Hi guys,
from the works done [here](https://github.com/icub-tech-iit/sw-management/issues/164) by Davide Tome, it seems that we need to specify both the core in order to update both.

In any case, I leave the floor to @marcoaccame, the expert of AMC board. 😺

traversaro:
> I will edit the PR accordingly.

Actually the files were already correct! As the mtb4c changes are actually already contained in the other PR as reported in https://github.com/robotology/robots-configuration/pull/693#issuecomment-2515074257, I think we can close the PR. The only doubt is that at this point why the update of the `amc` boards failed in https://github.com/robotology/icub-firmware-build/issues/176#issuecomment-2514964433, ideally we should try again now, but the whole process is a bit long so that may not be immediate, thanks!

valegagge:
Hi @traversaro , thanks!
I just realized that I opened an issue to take trace of this issue. see please here: https://github.com/icub-tech-iit/sw-management/issues/172.

I also added the reference to this PR, and we'll take care of this task as soon as possible.

pattacini:
Ok, to sum up, we will stick to #692, which should be merged before 2024.11.0 is out.

PR #700, opened 2024-12-18
fixed mechanichal values for the lego setup

SanLordKevin:
fixed mechanichal values for the motor gearbox and encoder to joint ratio, lego setup is simply a motor and an aea3


SanLordKevin:
cc @pattacini @marcoaccame

pattacini (approved):

PR #705, opened 2025-01-20
R1SN003: Move POS devices on top and calibrators on bottom, tune finger pids

mfussi66:
This PR aims to fix an issue in which the calibrator devices are not able to read the finger data if the latter start streaming data too late. By moving the former on top of the yarprobotinterface file, and the latter on bottom, this issue gets solved, like in ergoCub.

Additionally, the Position PIDs of the fingers are properly tuned.

cc @fbrand-new @randaz81

### DIFF
--- a/R1SN003/R1.xml
+++ b/R1SN003/R1.xml
@@ -4,6 +4,18 @@
 <robot name="R1SN003" build="1" portprefix="cer" xmlns:xi="http://www.w3.org/2001/XInclude">
 <devices>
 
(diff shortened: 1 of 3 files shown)


pattacini (changes requested):
Thanks for the fix @mfussi66!
Would you mind making it land on `master`, instead, since it is a fix?

Changing the base branch straight away doesn't work as it needs rebasing.

pattacini (approved):
Thanks heaps!

PR #723, opened 2025-03-25
Disable firmware check for amc2c board

LeoGori:
The robot is currently using a fw version greater than 3.0.4. 
This PR disables the firmware check, otherwise yarprobotinterface will not start.


pattacini (approved):
Given the rush, I can accept this modification.

However, it may be probably worth considering extending it to other boards and other R1 robots, as we did for ergoCub.

If you agree, I can deal with that.
cc @randaz81 

Merging in the meantime.

PR #745, opened 2025-06-27
Fix local-workflow-tech.md

S-Dafarra:
Setting the variable before the command does not seem to work. I quickly checked on Ubuntu and
```
var=5 echo ${var}
```
outputs empty, while
```
var=5
echo ${var}
```
outputs `5` as expected even without `export`


pattacini (approved):
Ok then!

PR #747, opened 2025-07-28
iCubGrenoble01/face: increase the low-level frequency to 1 kHz

Nicogene:
As per title


pattacini:
Hi @Nicogene 

I would limit the change to the face, exclusively.

PR #748, opened 2025-07-28
iCubGenova11 add files for fineCalibrationChecker

MSECode:
This PR adds file that are needed for using the fineCalibrationChecker on iCubGenova11
They are mainly useful for the PI demo but it might be useful to have them for further tests or for actually using the fine calibrator when we observe calibration errors with the robot


pattacini (changes requested):
Hi @MSECode 

CI failed with the following reports:
- [dry-run_log_iCubGenova11_xml.txt](https://github.com/user-attachments/files/21471976/dry-run_log_iCubGenova11_xml.txt)
- [dry-run_log_iCubGenova11_dev.txt](https://github.com/user-attachments/files/21471978/dry-run_log_iCubGenova11_dev.txt)

Also, there's one critical comment raised by @gemini-code-assist requiring your attention.

iCubGenova11/calibrators/left_arm-calib.xml
```diff
@@ -31,7 +31,7 @@
 			<param name="startupPosThreshold">    2          2          2          2         90     90     90    90    90    90    90    90    90     90     90    90    </param>
 		</group>
 
-		<param name="CALIB_ORDER">(0 1 2 3) (4) (5 6 7) (8 9 11 13) (10 12 14 15)</param>
+		<param name="CALIB_ORDER">(0 1 2 3) (4) (5 6 7) (8 9 11 13) (10 12 14)</param>
```
pattacini, line 34: @MSECode, was this change intended?

iCubGenova11/calibrators/right_arm-calib.xml
```diff
@@ -33,7 +33,7 @@
 </group>
 
 
-	<param name="CALIB_ORDER">(0 1 2 3) (4) (5 6) (8 9 11 13) (10 12 14 15)</param>
+	<param name="CALIB_ORDER">(0 1 2 3) (4) (5 6) (8 9 11 13) (10 12 14)</param>
```
pattacini, line 36: Same here.

pattacini (commented):
A couple of doubts regarding joint 15, @MSECode

iCubGenova11/calibrators/left_arm-calib.xml
pattacini: >Anyway I can remove that change from this PR and we can double check that directly on the robot.

Yep, thanks! I'd go with this solution.

pattacini:
> CI fails now while checking XML schema:

The schemas we're currently enforcing are available at:
- https://github.com/robotology/robots-configuration/tree/master/tests/dry-run/check-xml/schemas

cc @MSECode

iCubGenova11/calibrators/left_arm-calib.xml
martinaxgloria: I can confirm, they are disabled due to:

- https://github.com/robotology/icub-tech-support/issues/2178

pattacini: The fact that they're disabled locally to the robot doesn't imply that they should be upstream in the repo. Something to consider with care, I'd say.

pattacini:
Hi @MSECode 

You may want to fiddle with the XML schemas on your local system instead of relying on the CI itself.
You may also consider exploiting the GH Codespaces on your fork, which are backed up by a [docker](https://github.com/robotology/robots-configuration/tree/master/tests/dockerfiles).

The dry-run material is contained in https://github.com/robotology/robots-configuration/tree/master/tests/dry-run:
- Run `configure-tests.sh` to build the `check-xml` binary and configure all the robots.
- The executable `check-xml` is devoted to verifying the XML schemas stored in this repo. It may be that, aside from the schemas, the program needs some adjustment, depending on your use case.

pattacini:
I'll do that then, with the docker.

pattacini:
> I'll do that then, with the docker.

Done 👍🏻 
See https://github.com/robotology/robots-configuration/actions/runs/16649881946.

pattacini:
To be sure of the alignment of resources, we'd need a fake push to this PR.

pattacini:
The "dev" dry-run now fails with this error:

```console
qt.qpa.xcb: could not connect to display 
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.
```

which is obviously related to the use of graphics in a headless environment.
Options:
- We can install X and QT on the docker. Don't know if it will work on GH Action. Also, this will make the docker maintenance unnecessarily unwieldy.
- We develop the option `--dry-run` for the new plugin so that it can be used in this context. The option will be activated by the dry-run machinery.
- We rule out the plugin outright from the test.

---

That said, there's still an error with the "xml" checker as pointed out in https://github.com/robotology/robots-configuration/pull/748#issuecomment-3136277898.

pattacini:
Putting the PR back to draft because of https://github.com/robotology/robots-configuration/pull/748#issuecomment-3140623195.

pattacini:
https://github.com/robotology/icub-main/pull/1037 has been just merged. Thus, regenerating the Docker image via https://github.com/robotology/robots-configuration/actions/runs/17210854350.

pattacini:
> The `xml` dry-run is still failing needing further inspection ❌

To solve this, one should build and run locally the application https://github.com/robotology/robots-configuration/tree/master/tests/dry-run/check-xml.

MSECode:
> First off, we could check if the https://github.com/robotology/robots-configuration/blob/master/tests/dry-run/check-xml/schemas/wrapper.xsd schema can also cover the new service.

We added it in previous commits: https://github.com/robotology/robots-configuration/pull/748/files#diff-6fe0def77e76ee260725ce5da773ad315db9354899bd9a5a45dd7b7b062339c8

Does it make the CI working or the CI takes the schema only from master branch? If so we should probably add that commit on another PR

MSECode:
Hi @pattacini,
thanks to support by @martinaxgloria we have find out that the failing on the xml is due to the following reason:
- `checkCalibratorsWrappersRemappers` fails due to the fact that `rawValuesPublisherServer` does not have a remapper

Specifically the situation is the following:
- when this check is evaluated at this line: https://github.com/robotology/robots-configuration/blob/master/tests/dry-run/check-xml/src/check-xml.cpp#L138 we have actually a mismatch between the device defined in the files `*-rawval-nws_wrapper.xml` (e.g `right_arm-eb3-j0_3-mc`) and the expected one `right_arm-mc_remapper` 
- the device passed as target is taken from the `calibrators` file as shown by the following lines: 
    - the call of the method: https://github.com/robotology/robots-configuration/blob/master/tests/dry-run/check-xml/src/check-xml.cpp#L212
    - the target device: https://github.com/robotology/robots-configuration/blob/master/tests/dry-run/check-xml/src/check-xml.cpp#L200
    - the node selected: https://github.com/robotology/robots-configuration/blob/master/tests/dry-run/check-xml/src/check-xml.cpp#L193

Thus, an option of what to do now is to modify temporary the `check.xml` with this, quite naive, proposed change:

```
            if(ele.find("rawval-nws_wrapper.xml") != std::string::npos) {
                if(!loadXmlFile(robot_dir, ele, doc_wrapper)) return false;
                pugi::xpath_node action_startup = doc_wrapper.select_node("//action[@phase='startup']/param[@name='device']/text()");
                device = trim(action_startup.node().value());
                if(device != "") std::cout << part <<" - WRAPPER CHECK PASSED!" << std::endl;
                else {
                    std::cerr << part << " - WRAPPER CHECK FAILED!" << std::endl;
                    pass = false;
                }
                found = true;
            }
            else if(ele.find("wrapper.xml") != std::string::npos) {
                if(!loadXmlFile(robot_dir, ele, doc_wrapper)) return false;
```

where we add a temporary if clause to let the checks on the `nws_wrapper` pass until we finalize the definition of the remapper for the rawValuesPublisherServer, which is already planned. 

We can also think to add a new method for checking wrappers that does not have a related remapper, which thus have a device different from the target specified in the calibrators.

pattacini:
Thanks @MSECode and @martinaxgloria for the tests and the exhaustive report!

I would honestly favor the second solution:

>We can also think to add a new method for checking wrappers that does not have a related remapper, which thus have a device different from the target specified in the calibrators.

The reasons for that:
1. It's a more thorough approach that may come in handy in the future in similar situations.
2. This solution can stay thus integrated and we don't need to come back to remove the workaround.

tests/dry-run/check-xml/src/check-xml.cpp
```diff
@@ -183,6 +233,13 @@ bool checkCartesian(const std::string& robot_dir, std::vector<std::string>& vect
 }
 
 bool checkCalibratorsWrappersRemappers(const std::string& robot_dir, std::vector<std::string>& vectorAllFiles, bool& pass){
+     // save first list of unremapped devices to skip the check for wrappers/remappers
+    // we cannot check the single element inside the for loop by passing ele to method
+    // since we first need to fill the list of targets by looking in hardware/motorControl directory
+    // while the unramapped devices are inside wrappers/motorControl directory
```
pattacini, line 239:
```suggestion
    // save first list of unremapped devices to skip the check for wrappers/remappers
    // we cannot check the single element inside the for loop by passing ele to method
    // since we first need to fill the list of targets by looking in hardware/motorControl directory
    // while the unramapped devices are inside wrappers/motorControl directory 
```

tests/dry-run/check-xml/src/check-xml.cpp
```diff
@@ -133,6 +177,12 @@ bool checkWrappersRemappers(const std::string& robot_dir, std::vector<std::strin
         if(ele.find("wrappers/motorControl") != std::string::npos){
             if(ele.find("wrapper.xml") != std::string::npos) {
                 if(!loadXmlFile(robot_dir, ele, doc_wrapper)) return false;
+                if (std::find(vectorUnremapped.begin(), vectorUnremapped.end(), ele) != vectorUnremapped.end())
+                {
+                    std::cout << part << " - SKIPPED SINCE UNREMAPPED!" << std::endl;
```
pattacini, line 182: I think it's important to report which device is this.

pattacini:
Thanks @MSECode!
Awaiting the CI before merging.

pattacini (approved):

pattacini:
The CI reports:

```
left_arm - SKIPPED DEVICE left_arm-eb1-j0_3-mc SINCE UNREMAPPED!
```

Is `left_arm_eb1-j0_3-mc` correct?
I would have expected the name of the new device to show up here, or is that we skipped the entire sub-part?

cc @MSECode

MSECode:
> The CI reports:
> 
> ```
> left_arm - SKIPPED DEVICE left_arm-eb1-j0_3-mc SINCE UNREMAPPED!
> ```
> 
> Is `left_arm_eb1-j0_3-mc` correct? I would have expected the name of the new device to show up here, or is that we skipped the entire sub-part?
> 
> cc @MSECode

Oks, that is due to the fact  I added to the debug message the name of the device parsed from this lines: https://github.com/robotology/robots-configuration/blob/733e47570b4f965c7515d2d9afe565e6a535e26c/iCubGenova11/wrappers/motorControl/left_arm-mc-rawval-nws_wrapper.xml#L7-L9 (since it is the  device name commonly used by the other methods for comparing with the target taken from the calibrator file)
If it is much clear to the end user I can add the name we used for the device taken from this line: https://github.com/robotology/robots-configuration/blob/733e47570b4f965c7515d2d9afe565e6a535e26c/iCubGenova11/wrappers/motorControl/left_arm-mc-rawval-nws_wrapper.xml#L4

MSECode:
Locally, I've got the following output:

`left_arm - SKIPPED DEVICE left_arm-rawval_nws_yarp SINCE UNREMAPPED!` 
So I think it is fine now.

Here the full output for our test. If it is ok, I'll commit right away.

[dry_run_iCubGenova11_xml.txt](https://github.com/user-attachments/files/22004644/dry_run_iCubGenova11_xml.txt)

PR #750, opened 2025-08-27
Disable all fingers due to issues with tendons awaiting for support

MSECode:
This PR disables all hand/finger joints on iCubGenova11 since we have some problems with the tendons.
Therefore, in order to do not damage the motors it is safer to keep all of the finger disabled for now.


MSECode:
We already had the full commented line just below for both calibration orders:

- https://github.com/robotology/robots-configuration/pull/750/files#diff-aebf0e685e67f7ff7ad3defc4a36f1b65cc4f3b2ca81a290e9d7b9f93d1ef595L35

- https://github.com/robotology/robots-configuration/pull/750/files#diff-7402d044b359808af480483e52b2168b714a5545541f1055d7cfeae145984f21L37

PR #759, opened 2025-09-26
update amc,amcfoc lego and template

SanLordKevin:
update `amc`,`amcfoc` lego configuration files and fix comment of template file


SanLordKevin:
cc @marcoaccame

PR #766, opened 2025-11-28
ergoCubSN002: adjust fingers limits for the left hand

Nicogene:
The left hand has the mk5.2 coupling law:
- https://icub-tech-iit.github.io/documentation/hands/hands_mk5_coupling/?h=hand+mk#mk52-coupling-variables


Nicogene:
> Would it be possible to port this in [icub-tech-iit/robots-configuration@`devel-ergoCubSN002`](https://github.com/icub-tech-iit/robots-configuration/tree/devel-ergoCubSN002). I am afraid that any subsequent change on that file due to maintenance will cause a conflict that becomes hard to solve

I cherry-picked the commit in that branch (I had indeed to solve a conflict):
- https://github.com/icub-tech-iit/robots-configuration/commit/d7b69ce3aaf611a992e636cb68914d51faaf4820

PR #770, opened 2025-12-10
Fix configuration files for remapper in iCubGenova11

MSECode:
This PR fixes the configuration files for iCubGenova11 in relation to the `rawValuesPublisherRemapper`. This should fix the CI as well

### DIFF
--- a/iCubGenova11/icub_all_no_legs.xml
+++ b/iCubGenova11/icub_all_no_legs.xml
@@ -97,6 +97,7 @@
 
     <!-- FINE CALIBRATOR CHECKER DEMO and TEST -->
     <xi:include href="wrappers/motorControl/left_arm-mc-rawval-nws_wrapper.xml" />
+    <xi:include href="wrappers/motorControl/left_arm-mc-rawval-nws_remapper.xml" />
     <!-- <xi:include href="wrappers/motorControl/right_arm-mc-rawval-nws_wrapper.xml" /> -->
     <xi:include href="checker/fineCalibrationCheckerConfig.xml" />
     </devices>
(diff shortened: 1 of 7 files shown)


pattacini (approved):

MSECode:
So, in order to fix the CI, we needed to update the `check-xml` due to the following problem:
- when working with devices that are not of type `motorControl`, e.g the `rawValuesPublisherServer` or `rawValuesPublisherRemapper` devices, the pipeline of the CI that checks the wrappers and remappers cannot follow the standard path.
- Specifically, the main issue is related to the fact that the `target` that is used to subsequently [check the wrappers](https://github.com/robotology/robots-configuration/blob/353e78dda81ab2d88a1ccc51a77d495cedcb93c7/tests/dry-run/check-xml/src/check-xml.cpp#L188-L195) and [remappers ](https://github.com/robotology/robots-configuration/blob/353e78dda81ab2d88a1ccc51a77d495cedcb93c7/tests/dry-run/check-xml/src/check-xml.cpp#L199-L206) saved in the `wrappers/motorControl` folder is taken at the beginning of the pipeline as the value of the `target` node in the `calibrator` file as shown at [this line](https://github.com/robotology/robots-configuration/blob/353e78dda81ab2d88a1ccc51a77d495cedcb93c7/tests/dry-run/check-xml/src/check-xml.cpp#L259)
- therefore, if we'd like to check any wrapper and remapper that is not defined in the `calibrator` file, we needed to update the code of the `ctest` application

Thus, the main issue that made the CI failing was the fact that for `iCubGenova11` we introduced the raw values publisher wrappers and remappers in the `wrapper/motorControl` folder. However, those remappers are different from the ones defined in the `calibrator` file.
So, at the current status those wrappers are saved to a different list from the one of the `motor control` devices and at the end we perform a specific check for the _not motor control_ devices by checking that the specific remapper is correctly defined in the related wrapper.
For the future, we can improve the current `ctest` application in order to make it more robust by generalizing the wrapper check for any _not motor control_ device (considering that others might be introduced), adding also the check of the control boards defined in the remapper.

PR #784, opened 2026-02-26
Remove MAIS reference from ergoCub SN001 and SN003 configuration files

SanLordKevin:
As per title. I could not find those boards in the logic file and in the configuration files they were not used in `ergocub.xml` file.
cc @MSECode


pattacini (approved):
Nice catch @SanLordKevin!

PR #786, opened 2026-05-06
iCubGenova11 hard limits precise values updated from CAD (torso and legs).

ale-git:
We save these values also because they are important for absolute encoders fine calibration.


pattacini (approved):
Thanks heaps!

PR #788, opened 2026-06-03
Fix Pilsen ergoCubSN003

valegagge:
In this PR, I changed the configuration file to update:
 1. the current limits 
 2. PWM limits in calibration
 3. re-enable the right arm WBD configuration. @martinaxgloria @MSECode 
 
 The current limits have been updated given the motor characteristics:
 
image

image

 
In addition, during the calibration phase, we need to reduce the PWM limit to `8000`.

The reasoning is as follows:

- The maximum current is `26 A`, but we use `20 A` as a safety margin.
- The motor resistance is approximately `0.4 Ω`, so:

  `20 × 0.4 = 8 V`

- Since the input voltage is approximately `40 V`, rescaling this value to our PWM unit gives `8000`.

Please note that the raw `MAXPWM` value is `32000`.

cc @fgarini @pattacini

### DIFF
--- a/ergoCubSN003/calibrators/torso-calib.xml
+++ b/ergoCubSN003/calibrators/torso-calib.xml
@@ -33,7 +33,7 @@
                                                                                                    
 <param name="startupPosition">                      0                       0                      0                       </param>
 <param name="startupVelocity">                      10                      10                     10                      </param>
-<param name="startupMaxPwm">                        5500                    16000                  5500                    </param>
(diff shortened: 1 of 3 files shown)


pattacini:
This PR addresses https://github.com/robotology/mesh-support/issues/2324#issuecomment-4611327939.

pattacini (approved):
Thanks!

PR #792, opened 2026-07-23
[iCubGenova11]: Contributing upstream

pattacini:
As per title.


pattacini:
Hi @martinaxgloria 

This [line](https://github.com/mesh-iit/robots-configuration/blob/devel-iCubGenova11/iCubGenova11/CMakeLists.txt#L26) is causing the CI to fail: https://github.com/robotology/robots-configuration/actions/runs/30019648886/job/89248715840#step:4:79.

```cmake
#Install the WalkingJoypad extra files
yarp_install(DIRECTORY extra/JoypadModule/dcmWalkingJoypad DESTINATION ${ICUBCONTRIB_CONTEXTS_INSTALL_DIR})
```

Is there any specific reason as to why we stick to `{ICUBCONTRIB_CONTEXTS_INSTALL_DIR}` rather than `${ICUBCONTRIB_ROBOTS_INSTALL_DIR}/${appname}`?

It seems that `{appname}` is missing anyway.

pattacini:
Also, I already had to commit to `devel-iCubGenoa11` to fix a conflict.
Thereby, we ought to pull the latest branch locally to he robot.

martinaxgloria:
Hi @pattacini,

I just copied the original CMakeLists [installation path](https://github.com/gbionics/walking-controllers/blob/1588924d04f6579005991caa4f0710b41947438a/src/JoypadModule/app/CMakeLists.txt#L5), I didn't know there would be some issues in doing the same way. However, I think that we should safely rely on `${ICUBCONTRIB_ROBOTS_INSTALL_DIR}/${appname}`, I don't see any issue with that.

pattacini:
We decided to remove those conf files as they're not robot specific.

PR #793, opened 2026-08-26
Cleanup for ergoCubSNxxx remappers

randaz81:
This PR is a cleanup, prepared by a script, for ergoCub remapper files, with no functional changes. 
It's valid for both YARP 3.x and 4.x series.

In these modified remapper files, the joint numeric indices (deprecated syntax, will be removed soon) have been replaced by joint names, and the parameter `joints` (number of joints) has been removed because it is no longer used.


pattacini:
Thanks @randaz81 for the cleanup.
However there are two aspects to consider before merging:
- CI is failing on the XML checker part. Probably, we need to fiddle with the checker somehow.
- Given that ergoCubSN003 is under integration test and is supposed to be delivered back to the client soon, I would rather postpone this PR after we complete the robot maintenance.

Nicogene (approved):
I would do a test on our SN001 before merging but LGTM!

randaz81:
> * CI is failing on the XML checker part. Probably, we need to fiddle with the checker somehow.

This is because the current .xsd schema is incorrect (I checked the CI artifact). 
I fixed it in the last commit. However,  I expect it to fail again, with ergoCub files now ok, and all the other iCub robots failing. I am now proceeding with two separate PRs: one for all the iCubs and one for the R1s.

> * Given that ergoCubSN003 is under integration test and is supposed to be delivered back to the client soon, I would rather postpone this PR after we complete the robot maintenance.

ok for me!

valegagge:
Hi @randaz81 , I relaunch the CI but it continues to fail on ergocubSN00 and ergocubSN003, while on ergocubSN002 it runs succefully.

image

Please, Can you fix it? thanks heaps

valegagge:
ergoCubSN001 and ergoCubSN003 fixed here: https://github.com/robotology/robots-configuration/actions/runs/33976813159/job/101334849353.

Thanks @pattacini

PR #794, opened 2026-08-27
Cleanup for iCubXXX remappers 

randaz81:
Should be merged only after: https://github.com/robotology/robots-configuration/pull/793.


pattacini (changes requested):
Hi @randaz81 

I've rebased this fix branch on top of `devel`.
R1SN001 and R1SN002 still fails the XML checks.

Could you please update these robots' files as well?

pattacini:
> R1SN001 and R1SN002 still fails the XML checks.

We will be updating only the R1 at Erzelli, while we will move the other to the icebox.

pattacini:
@randaz81 we need to merge this PR urgently as @SanLordKevin has commits to add for iCubGenova02.

